PR #9, opened 2025-06-29
Adding libraries page and data simplification

TerminalFi:
https://github.com/user-attachments/assets/0b417ed1-6524-470b-89b7-ce0747e18a5e


braver:
If it helps, I could also live without sorting options on the library page. There are fewer useful options here, and likely you're never comparing multiple alternatives for the same feature: you just need to be able to find the specific one you're looking for.

kaste:
Don't know where this fits but seeing the results *lists* here feels more useful and superior.  Typically, the landing page could have cards, even visually very dense (i.e. with less information) cards, but as soon as I search -- I think -- scanning a list is optimal.  It is just eyes on the left, then top -> down.  That one doesn't fit at all here:  Sort by last updated (most recent) and last added are useful.  Sort by stars reversed typically not so.  :sorry:

braver:
> but as soon as I search -- I think -- scanning a list is optimal. It is just eyes on the left, then top -> down

Yeah, that’s a great way to do it: cards on the homepage, list when you search. 👍🏻

MarketHubb:
When viewing items in a vertical list, independently encapsulated cards with a vertical gap is (probably) too much structure. 

I would suggest that, when in a list view, applying background contrast the _container_ and bifurcating the individual packages with a simple border-bottom. 

We're already using a grid layout on the parent `<ul>`, so as long as there's some targetable class on the body or list container denoting the layout type (grid or list) you could easily modify the existing card layout to become a list layout with a descendent selector (almost no modification to the card structure or existing css).

I took the existing layout in my [PR](https://github.com/packagecontrol/thecrawl/pull/10) from last night, and simply wrapped the package title and author in a flex container so they could output in the same row on list.

**Example**

Screenshot 2025-06-30 at 10 00 05 AM

braver:
Love how that looks! 👍🏻

FichteFoll:
Just for the record and to provide some context to this PR, PC5 may or may not have a fallback to the entirety of pypi, which would make a separate library listing redundant. (via https://github.com/wbond/package_control/pull/1705#issuecomment-3012636244)

However, PC 5 is still far and away, so don't let this stop you from adding a libraries page.

braver:
@TerminalFi given that the dataset is indeed just 119 records, with no metadata or detail pages, and the target audience is much more niche... Let's just add a single static page with libraries. That way we can build that listing bespoke for this purpose without introducing factories and abstractions, and without pulling all the complexities into this space as well. A simple keyword search can be added if someone feels like implementing that (honestly for this use case I think CTRL+F also works), but we don't have to mess with sorting too much (most of the metadata to sort on isn't here anyway).

Given that brief, do you still feel like implementing it and do you have time for that, or shall I give it a go?

@deathaxe @FichteFoll am I correct in understanding that this would replace http://packagecontrol.github.io, and that that's built on effectively the same [data](https://raw.githubusercontent.com/packagecontrol/channel/main/repository.json)?

deathaxe:
Old page just pulls static list of libraries and shildio badges to provide current versions, client-side. Given its source is the old channel, it is more or less out-dated anyway. With a libraries page added to packages.sublimetext.io, I'd vote to retire the old page.

For the new page however, we should probably use workspace.json to also display resolved versions. There may be different ones for python 3.3 and 3.8 and maybe 3.13 later on.

deathaxe:
Sure.

PR #10, opened 2025-06-30
Modernize package grid design (light)

MarketHubb:
This PR contains:

**Opinionated design changes to the grid in light mode**

- Inverted body and card background colors to improve card contrast. Using a (light) gray body background has the added benefit of making input fields, like the search bar, stand out.
- Reduced amount of bolded text in layout, while increasing text color contrast (as the saying goes: when everything is bold, nothing is bold)
- Adjusted package tags to align left. I imagine most people who use this site will read left to right. As such, having the _first_ badge/pill appear in the same position on each card should significantly improve the users ability to scan the list (see screenshots for comparison) 
- Soften the layout with rounded borders / badges and slightly increase padding and gap

**Semantic css vars + introduce utilities.css**

-  These can be removed if the team is opposed, but if we're going to be using home-grown css vars instead of a framework, it might helpful to be semantic in our naming convention. Currently we have items like `--background: #fff;` whereas something as simple as `--white: #fff;` would be more clear (especially when considering things like light/dark mode will change what color is associated with "background")
- I took the already defined "foreground" and created a semantic `--gray` var, then using pre-defined (HSL) stops, added a handful of light/dark semantic vars to show an example of what we could do based on whatever our defined color palette is. 

**Detailed list of changes:**

  - Add gray-scale color palette for better design consistency
  - Update card components with white backgrounds, rounded corners & left-aligned tags
  - Convert typography units from px to rem for better scalability
  - Implement 3-column grid layout with increased spacing between cards
  - Add utilities.css for design tokens (spacing, font-weights) and future (non typography-related) vars
  - Update label buttons with softer styling and pill-shaped borders
  - Improve text contrast and readability with color-mix functions
  - Change site background to lighter gray for better card contrast
  
**Before:**
before

**After**
after

### DIFF
--- a/static/style/colors.css
+++ b/static/style/colors.css
@@ -4,6 +4,12 @@
   --primary-accent: #fd971f;   /* orange */
   --secondary-accent: #66d9ef; /* blue */
   --tertiary-accent: #a6e22e;  /* green */
+  --white: #fff;
+  --gray-lightest: #f9fafb;
+  --gray-light: #CDD7D9;
+  --gray: #586E75;
+  --gray-dark: #38464B;
+  --gray-darkest: #262F32;
 
   --primary-accent-darker: #4694A3;
 
@@ -11,6 +17,7 @@
   --background-slightly-darker: #fcfcfc;
   --background-darker: #f2f2f2;
   --background-darkest: #cacaca;
+
 
   --foreground: #586e75;
   --foreground-darker: #38464b;
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -1,14 +1,17 @@
 .grid {
   display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(22rem, 1fr));
-  gap: 11px;
+  grid-template-columns: repeat(3, minmax(0, 1fr));
+  gap: calc(var(--spacing) * 4);
   margin: 0;
   padding: 0;
-  > * {
+  align-items: stretch;
+
+  >* {
     display: flex;
     margin: 0;
     padding: 0;
-    > * {
+
+    >* {
       flex: 1;
     }
   }
@@ -17,9 +20,11 @@
 .card {
   display: flex;
   flex-direction: column;
-  background: var(--background-slightly-darker);
-  border: 1px solid var(--background-darker);
-  padding: 1rem;
+  background: var(--white);
+  border: 1px solid;
+  border-color: color-mix(in oklab, var(--gray-light) 50%, transparent);
+  padding: 1.25rem 1.5rem;
+  border-radius: 6px;
   position: relative;
   transition: border-color .3s ease-in-out;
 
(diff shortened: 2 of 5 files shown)


MarketHubb:
Absolutely I can whip up a dark-mode variant. Before I do however, help me understand the how we intended to group the collections of packages on the new site.

**Current paradigm**

For example, on the current packagecontrol.io site the homepage displays the following 4 collections:

- Trending
- New
- Popular 
- Label

The "Browse" page displays 6 collection. Each of the homepage collections, plus:

- Author
- Updated

If we intend to mimic that, I would strongly advise _against_ a vertically stacked, single column list like we currently have. These are inherently "scan" pages and we would want to reduce the amount of vertical scrolling.

**Options for new site (if we intended to display the same collections**

- Mimic the current packagecontrol.io layout, but modernize
- Incorporate dynamic layouts (tabs, pills, accordions, collection badges)

browse
home

braver:
A related discussion going on in #9. See https://github.com/packagecontrol/thecrawl/pull/9#issuecomment-3018644226

So, we don't yet have all the data the old site has (notably usage stats). But the goal is similar. Like you say: scannable overview grids for those initial pages. Then as you start searching, especially if you're searching by name (ie. not filtering), then it becomes beneficial to have a list.

It's not necessarily a goal to match the old site in "richness" though. I'm not sure we'll have the data, or the immediate desire, to add "popular" and "trending" sections.

TerminalFi:
@MarketHubb it might be nice to use the same color scheme from docs.sublimetext.io

TerminalFi:
> A related discussion going on in #9. See [#9 (comment)](https://github.com/packagecontrol/thecrawl/pull/9#issuecomment-3018644226)
> 
> So, we don't yet have all the data the old site has (notably usage stats). But the goal is similar. Like you say: scannable overview grids for those initial pages. Then as you start searching, especially if you're searching by name (ie. not filtering), then it becomes beneficial to have a list.
> 
> It's not necessarily a goal to match the old site in "richness" though. I'm not sure we'll have the data, or the immediate desire, to add "popular" and "trending" sections.

https://stats.sublimetext.io/all-events

TerminalFi:
> 
> > it might be nice to use the same color scheme from docs.sublimetext.io
> 
> That does have an orange and a greyscale that would work, but otherwise it looks a bit limited.

Primarily speaking about the primary BG colors

MarketHubb:
> > > it might be nice to use the same color scheme from docs.sublimetext.io
> > 
> > 
> > That does have an orange and a greyscale that would work, but otherwise it looks a bit limited.
> 
> Primarily speaking about the primary BG colors. Even sublimetext.io should probably align

The layout in the screen shot is using the (already defined) gray's in the `colors.css` as the "base" and shading off of them. However, I much prefer the base colors in the [docs](https://docs.sublimetext.io/) site, and can easily whip that up.

That said, since we've moved to this new repo with distinct branches, and this hasn't been merged (and now has conflicts), should I fetch the latest and just submit a new PR with dark and light? What's the best path forward?

braver:
Yeah, a bunch of stuff has moved around in the mean time. 

I am not really good at colors so if you want to take a stab at redoing the color scheme based off the documentation site, I'd be much obliged. If it turns out the shade of orange is off, let me know if you can work with the logo SVG, or I need to go back to the original in Affinity Designer.

braver:
Thanks all, I took a lot of ideas and inspiration from this discussion and these proposals, and redid the color scheme, renamed most variables, and restyled various elements. It's now a much closer match also to the [docs](https://docs.sublimetext.io/) site, which makes the project as a whole feel more coherent. This just shipped via #58

Issue #5, opened 2025-06-29
Link to the rendered readme on GitHub


braver:
Original report: https://github.com/packagecontrol/the-packages-site/issues/11

The link is currently to the raw readme, ie. Markdown without clickable links or visible images. 

We might want to put in the effort of trying to render the readme (including images, build status label, and all the challenges related to that). Or try to "find" the readme URL on GitHub that can be seen in all its glory there in the context of the repo.

kaste:
I do like the glory part but readme is also a key allowed in the registry.  E.g. wbond never used GitHub and links to some arbitrary urls iirc.

kaste:
From the spec:

```
        # mandatory keys but with null or empty defaults
        "description": pkg.get("description"),
        "previous_names": pkg.get("previous_names", []),
        "labels": pkg.get("labels", []),
        "readme": pkg.get("readme"),
        "issues": pkg.get("issues"),
        "donate": pkg.get("donate"),
        "buy": pkg.get("buy"),
```

That means, the keys are always present -- (why?) -- *but* can be null (or an empty array).

PR #13, opened 2025-06-30
Add rendered README

kaste:
This implements a client side renderer.   Rel #5

### DIFF
--- a/packages/package.njk
+++ b/packages/package.njk
@@ -91,20 +91,29 @@ if (cached && (now - cached.time) < ttl) {
   target.innerHTML = cached.html;
 } else {
   fetch('{{ pkg.readme }}')
-    .then(res => res.text())
+    .then(res => {
+      if (!res.ok) throw new Error(`HTTP error ${res.status}`);
+      return res.text();
+    })
     .then(md => {
       if (DOMPurify.isSupported) {
-          const html = marked.parse(md);
-          const safe = DOMPurify.sanitize(html);
-          sessionStorage.setItem(cacheKey, JSON.stringify({ html: safe, time: now }));
-          target.innerHTML = safe;
-        } else {
-          const escaped = md
-            .replace(/&/g, '&amp;')
-            .replace(/</g, '&lt;')
-            .replace(/>/g, '&gt;');
-          target.innerHTML = escaped;
-          target.className = 'raw';        }
+        const html = marked.parse(md);
+        const safe = DOMPurify.sanitize(html);
+        sessionStorage.setItem(cacheKey, JSON.stringify({ html: safe, time: now }));
+        target.innerHTML = safe;
+      } else {
+        const escaped = md
+          .replace(/&/g, '&amp;')
+          .replace(/</g, '&lt;')
+          .replace(/>/g, '&gt;');
+        target.innerHTML = escaped;
+        target.className = 'raw';
+      }
+    })
+    .catch(err => {
+      console.error('Failed to load readme:', err);
+      target.innerHTML = '"😒, the readme failed to load."';
+      target.className = 'raw';
     });
 }
 </script>
--- a/packages/package.njk
+++ b/packages/package.njk
@@ -74,3 +74,29 @@ eleventyComputed:
   </li>
   {% endif %}
 </ul>
+
+<script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/dompurify/dist/purify.min.js"></script>
+
+<div id="md"></div>
+
+<script>
+const cacheKey = `md:{{ pkg.readme }}`;
+const ttl = 60 * 60; // 1 hour in seconds
+
+const target = document.getElementById('md');
+const cached = JSON.parse(sessionStorage.getItem(cacheKey) || "null");
+const now = Math.floor(Date.now() / 1000);
+if (cached && (now - cached.time) < ttl) {
+  target.innerHTML = cached.html;
+} else {
+  fetch('{{ pkg.readme }}')
+    .then(res => res.text())
+    .then(md => {
+      const html = marked.parse(md);
+      const safe = DOMPurify.sanitize(html);
+      sessionStorage.setItem(cacheKey, JSON.stringify({ html: safe, time: now }));
+      target.innerHTML = safe;
+    });
+}
+</script>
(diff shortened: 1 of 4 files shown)


kaste:
image

braver:
> IMO better than packagecontrol.io

A thousand times better! 🏅

braver:
LOL, your screen is really large :) But yeah, the header should be contained to some kind of or reasonable max, and then perhaps just left-align things.

packages/package.njk
```diff
@@ -74,3 +74,29 @@ eleventyComputed:
   </li>
   {% endif %}
 </ul>
+
+<script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/dompurify/dist/purify.min.js"></script>
+
+<div id="md"></div>
+
+<script>
```
braver, line 83: Can we move this (importing the dependencies) to a JS file? We'll need to pass the readme URL of course... I don't know, this is pragmatic of course, but it is preferable to leave the JS out of the HTML mostly I would say.

static/styles.css
```diff
@@ -105,3 +105,15 @@ dl {
     }
   }
 }
+
+div#md {
+    border-top: 4px solid #eee;
+    margin-top: 1em;
+    padding: 1.5em;
+    background: #f9f9f9;
```
braver, line 113:
```suggestion
    border-top: 4px solid var(--background-darker);
    margin-top: 1em;
    padding: 1.5em;
    background: var(--background-slightly-darker);
```

packages/package.njk
```diff
@@ -74,3 +74,29 @@ eleventyComputed:
...
+const target = document.getElementById('md');
+const cached = JSON.parse(sessionStorage.getItem(cacheKey) || "null");
+const now = Math.floor(Date.now() / 1000);
+if (cached && (now - cached.time) < ttl) {
+  target.innerHTML = cached.html;
+} else {
+  fetch('{{ pkg.readme }}')
+    .then(res => res.text())
+    .then(md => {
+      const html = marked.parse(md);
```
braver, line 96: Should we set the [gfm option](https://marked.js.org/using_advanced#options)? Assuming most other services (GitLab et al) also use GitHub Flavored Markdown?

kaste:
> LOL, your screen is really large
Changes all the time.  amazon sets 1500px as max-width right now.

packages/package.njk
kaste: Typically, import the dependencies in the header.  This way they preload on the main page.  Ideally we had different headers, but I did not find that.  E.g. on the main page wehere they're not used, just preload.  But the details page we require.  KISS though.  We also need to actually download them locally and deploy our versions.  Otherwise, we get new versions... didn't find that either as we already have minisearch as a dependency.

braver: > Typically, import the dependencies in the header.

If they provide an esm module that’d be ideal and can be imported straight in the script without adding script tags.

But indeed it’s small and encapsulated this way, I don’t really mind.

packages/package.njk
```diff
@@ -74,3 +74,29 @@ eleventyComputed:
...
+const ttl = 60 * 60; // 1 hour in seconds
+
+const target = document.getElementById('md');
+const cached = JSON.parse(sessionStorage.getItem(cacheKey) || "null");
+const now = Math.floor(Date.now() / 1000);
+if (cached && (now - cached.time) < ttl) {
+  target.innerHTML = cached.html;
+} else {
+  fetch('{{ pkg.readme }}')
+    .then(res => res.text())
```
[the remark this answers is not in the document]
braver: Can we simply print “Sorry, the readme failed to load”, instead of the readme? That’s both low effort and easier to spot without going into the console.

braver:
> Why there are so many files right now, tiny css and js files, but no "build" step? minisearch isn't in the head too. I don't know when it is loaded even. T.i. the exact timing of all of it.

That’s pretty standard these days. You can import dependencies in your script when you’re use ES Modules (esm), even when you’re using them from a CDN. The browser figures out what’s needed and when to load stuff. Many small files is no longer an issue for load times, only when you go over 200 or so does it start to make sense to introduce build steps. So you can organize your native code with smaller modules like you would if you did have a build step.

braver:
Remember you’re still looking at something in the MVP stage that’s as simple as it can be, not something hyperoptimized 😉

packages/package.njk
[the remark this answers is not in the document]
kaste: Done.  Also had to add a fallback if DOMPurify is not available.  In this case I just render the markdown in a `pre` block.

packages/package.njk
```diff
@@ -74,3 +74,45 @@ eleventyComputed:
...
+          const safe = DOMPurify.sanitize(html);
+          sessionStorage.setItem(cacheKey, JSON.stringify({ html: safe, time: now }));
+          target.innerHTML = safe;
+        } else {
+          const escaped = md
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;');
+          target.innerHTML = escaped;
+          target.className = 'raw';        }
```
braver, line 110:
```suggestion
        const html = marked.parse(md);
        const safe = DOMPurify.sanitize(html);
        sessionStorage.setItem(cacheKey, JSON.stringify({ html: safe, time: now }));
        target.innerHTML = safe;
      } else {
        const escaped = md
          .replace(/&/g, '&amp;')
          .replace(/</g, '&lt;')
          .replace(/>/g, '&gt;');
        target.innerHTML = escaped;
        target.className = 'raw';
      }
```

Just some whitespace. Nice though 👍🏻

kaste:
It is just a non argument here.  You made an *abstraction*, and that's a cost of it.   Just add preconnect, prefetch to the header for the fonts and the scripts.  That feature came at the same time as http2.  And test with 4g slow.   It is the WWW and not the web of the 35 richest countries.  :grin:

braver:
> It is the WWW and not the web of the 35 richest countries.

Fair point.

Just let's optimize at some later point shall we?

packages/package.njk
kaste: Damn, I f** up the white-space.  2 spaces aren't enough for me.

braver:
I’ll probably merge the whole shebang tomorrow morning. It’s so cool that you were able to make this, to this level. I really thought it would be a super nasty can of worms. Guess I didn’t consider that whatever does the work in the old site must also be kind of outdated by now.

PR #14, opened 2025-06-30
Readme tweaks

braver:
- Wider main width
- Move readme rendering script into its own script file
- Load dependencies via esm imports

In terms of parallelization and performance...

Before:
Scherm­afbeelding 2025-06-30 om 20 16 32

After:
Scherm­afbeelding 2025-06-30 om 20 16 13

Of course every time I reload the timing is slightly different. IMO it's close enough and fast enough that it doesn't matter.

I do prefer the separation of concerns and moving the scripts out of the markup.

I made a provision that the readme section can be left empty and it won't have a border or take up space. I did not implement error handling in the fetching of the readme.


kaste:
If it is already ideal 
image
 Why change it?

braver:
> Why change it?

Because there is a pattern established here where the scripts are in one place and markup is in another place, and all dependencies are loaded the same way. That will make it easier for us to, e.g., create a build step and improve performance for the other not-richest countries ;) And just consistency.

PR #15, opened 2025-06-30
Refactor Minisearch usage

kaste:
* Remove `Data` abstraction.
* Use a global, singleton minisearch instance to avoid lexing/splitting the data per search
* Use Minisearch's advanced search feature to search for "author" et.al.
* Use its results directly
* Allow the simpler `author:kaste` (instead of `authot:"kaste"`) if the name does not contain spaces.  
* Same for label and platform


kaste:
image

static/module/search.js
```diff
@@ -1,91 +1,81 @@
...
+const data = await fetchSearchData();
+const minisrch = new minisearch({
+  idField: 'name',
+  fields: ['name', 'description', 'author', 'platforms', 'labels'],
+  // keep in sync with searchindex.json.njk
+  storeFields: [
+    'name', 'description', 'author', 'stars', 'platforms', 'labels', 'permalink'
+  ],
+  searchOptions: {
+    boost: { author: 2 },
```
braver, line 18: Shouldn’t the package name get boosted instead? Or am I missing something?

kaste: One would think so, but names are a bit special, esp. with fuzzy.  If you start typing "kast", you otherwise get "paste" results first.  Or you type "odat" but get "oat".  Names are not normal words so there is no overlap or fighting with terms usually.

braver: Yeah I get that, and I search for my packages as I’m sure you search for yours. But I can also explicitly filter and get exact results. When I search for a language though, many of which have only a few characters in their name, I would prefer if we limit the number of false positive hits in authors or even descriptions, and prefer prioritized hits on the package name.

braver: That said, I haven’t tried it yet, so maybe this is indeed better, just asking about the reason for changing the boost.

braver: Ok yeah, this works really well in fact. I was worried that with something like `erb` matches on authors would get it in the way, but it's really good actually.

Screenshot 2025-07-01 at 14 57 39

FichteFoll:
> * Allow the simpler `author:kaste` (instead of `authot:"kaste"`) if the name does not contain spaces.

I was about to request this since I also ran into it on Monday but decided to go through my notifications and the recent developments first. Duplicate avoided :+1:

PR #16, opened 2025-07-01
markdown fixups

kaste:
* We actually supported other text variants. .creole, .textile, and .rst.  Just render them verbatim.
* Store verbatim results in the session as well
* resolve relative URL's


kaste:
Ah, that moved.  Damn.  Let's rebase or so.

braver (approved):

PR #17, opened 2025-07-01
For the blind people

kaste:
GitHub sets this to 16px even.  Old packagecontrol.io had 15px.

### DIFF
--- a/static/style/typography.css
+++ b/static/style/typography.css
@@ -64,7 +64,7 @@ h3 {
 
 p,
 ul {
-  font-size: 14px;
+  font-size: 15px;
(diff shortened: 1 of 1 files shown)


braver:
16 is good. Changing that affects how the layout works, so it'll need a bit more than just a bump. But I don't mind going for something that's a bit more spacious all around. I'll have a look!

braver:
merged via 129734fbd964579e818183f7e287470dec478e96

braver:
Made everything a bit bigger based on 16px base font-size. It is better! My small laptop screen fooled me into making things too tiny.

PR #21, opened 2025-07-01
Omit the timezone from the build info in the footer

### LINKED ISSUE #19: Reduce timestamp noise in footer
Image

The line above already mentiones that all times are in UTC.

kaste:
Fixes #19


braver (approved):
🚀

Issue #25, opened 2025-07-01
Should auto-replace references to `packagecontrol.io` with `packages.sublimetext.io`


TerminalFi:
![Image](https://github.com/user-attachments/assets/af57acc9-e547-4681-81ae-f404a03476e5)

Issue #23, opened 2025-07-01
Permalinks should be in packages (plural)


braver:
Compare, our URL
```
https://packages.sublimetext.io/package/Evergarden%20color%20schemes/
```

And the old site
```
https://packagecontrol.io/packages/Evergarden%20color%20schemes
```

Ideally we match that, so that if we do ever get hold of the old domain name (or it redirects to us) Google results, people's bookmarks, and potentially links in README's etc., all carry over transparently.

PR #26, opened 2025-07-01
replace packagecontrol.io and also fix permalinks to be packages

TerminalFi:


TerminalFi:
fixes #25 and #23

kaste:
I think this is good, the readme stuff is aggressive but we'll see.  Likely this will hit us when they link to the pc.io documentation.  (Never tried: how big is the corpus of all readme's we have here?  Can this be fetched server side without breaking it up into n tasks that fail in m ways?)  If we had the corpus we could answer that question what links to pc.io we actually expect here.

PR #29, opened 2025-07-01
Safely access optional package information

kaste:
Rel https://github.com/packagecontrol/thecrawl/issues/5#issuecomment-3018844970

### DIFF
--- a/packages/package.njk
+++ b/packages/package.njk
@@ -7,13 +7,15 @@ pagination:
   alias: pkg
 eleventyComputed:
   title: "{{ pkg.name }}"
-  description: "{{ pkg.description }}"
+  description: "{{ pkg.description | default('') }}"
 ---
 
 {% import "packages/macros.njk" as pack %}
 
 <h1>{{ pkg.name }}</h1>
+{% if pkg.description %}
 <p class="description">{{ pkg.description }}</p>
+{% endif %}
 
 <p>
   By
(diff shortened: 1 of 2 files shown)


packages/package.njk
```diff
@@ -7,13 +7,15 @@ pagination:
   alias: pkg
 eleventyComputed:
   title: "{{ pkg.name }}"
-  description: "{{ pkg.description }}"
+  description: "{{ pkg.description | default('') }}"
 ---
```
kaste, line 11: Q: where is this used?

packages/package.njk
```diff
@@ -57,12 +59,16 @@ eleventyComputed:
     <a class="button" href="{{ pkg.homepage }}">{{ pack.logo('home') }} <span>Website</span></a>
   </li>
   {% endif %}
+  {% if pkg.details %}
```
kaste, line 62: E.g. wbond SFTP has not details as it is proprietary.

packages/package.njk
braver: Meta and OpenGraph descriptions in the head include file.

kaste: Ah.  So that's also basically how you push content to the head section.   E.g. scripts etc.

braver: > But then you must use "extends" I think.

Correct. That would in general be the preferred approach anyway. Currently it doesn't matter much, but it would clean things up a little and allow us to use blocks. Which might come in handy when we add libraries or other kinds of content.

PR #30, opened 2025-07-01
`scrollIntoView` only if the target is not visible

kaste:
Two variants, (1) completely not visible or (2) partly not visible. I choose (1).

To be clear: if the "h1" is partly visible, I do *not* scroll.


TerminalFi (approved):

braver:
I ended up moving some stuff into the list class, seemed like a neat place to put all of the "search results lists" related _stuff_. With goSearch in there it's also no longer needed to have a global. At least... it seems to all still work 🤞🏻 d99aa36adef1d885787c65dd295c5797174b6259

PR #31, opened 2025-07-01
Transform raw readme links to its glory

kaste:
Rel #5

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -40,11 +40,44 @@ module.exports = function (eleventyConfig) {
   // eslint-disable-next-line no-unused-vars
   const live_packages = Object.entries(data.packages).map(([id, pkg]) => pkg).filter(pkg => !pkg.removed);
 
+  // if readme is in pkg
+  // transform some links
+  // https://raw.githubusercontent.com/relikd/CUE-Sheet_sublime/main/README.md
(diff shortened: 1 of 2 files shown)


.eleventy.js
```diff
@@ -40,11 +40,44 @@ module.exports = function (eleventyConfig) {
...
+        // GitHub raw to blob
+        readme_url = readme_url.replace(
+          /^https:\/\/raw\.githubusercontent\.com\/([^/]+)\/([^/]+)\/([^/]+)\/(.+)$/,
+          'https://github.com/$1/$2/blob/$3/$4'
+        );
+        // GitLab raw to blob
+        readme_url = readme_url.replace(
+          /^https:\/\/gitlab\.com\/([^/]+)\/([^/]+)\/-\/raw\/([^/]+)\/(.+)$/,
+          'https://gitlab.com/$1/$2/-/blob/$3/$4'
+        );
```
TerminalFi, line 68: Gitlab sadly supports many nested project / groups. For example the following is a valid Gitlab URL where `trmnl` `test` `test` are all groups and then `thecrawl` is a project

Valid -> `https://gitlab.com/trmnl/test/test/thecrawl/-/raw/main/README.md`

braver (approved):
Regexes are always kinda scary but looks ok. Even with so many packages it’s a relatively small domain anyway.

PR #33, opened 2025-07-01
Sort releases by relevance

### LINKED ISSUE #18: Sort versions by latest/highest target build first
Image

The most relevant version should be the first one.

kaste:
Newest first, if same date newest sublime specifier first.

Fixes #18


kaste:
image
image

kaste:
On a different planet: I think platforms are actually "Windows", "Linux", "Mac" or "macOS" or "OSX".  "Mac" is okay.

kaste:
On the other planet:  I just wanted to compare "windows" => "Windows" et.al. And for "macos", "MacOS"/"Macos" is untypical and "macOS" would be standard, and then just "Mac" would also have yadayada a leading uppercase character.  🙄

braver:
Yeah, ok ... I'm afraid this is the planet we got. Which is sad on many levels 😂

braver:
It really does.
So, shall we correctly capitalise the platform labels then? Makes sense do it.

kaste:
Sure, I did no CSS at all, just the tag.  Can be moved around too.

PR #38, opened 2025-07-02
`search` type for search input

michaelblyons:
Signal to client browsers that this is the search for the site.

It may only change the "Go" in mobile keypads to a :mag:, but I think some browsers can give you a custom search engine for the site.

Apologies if this botches the CSS. I didn't build out the site.


braver:
> Apologies if this botches the CSS. I didn't build out the site.

No it's fine, the type is not targeted specifically. Thanks for this, I wasn't aware of that input type (I do this stuff for a living but can't for the life of me keep all the input types straight in my head 😅).

PR #43, opened 2025-07-04
Fix dark theme flicker bug

kaste:


kaste:
Throttling is here:
image

braver:
Thanks again! I just reran, and I finally got a reproduction. With network throttled to 4G and 20x cpu throttling. I guess I run a much too privileged first world setup here 😅 It's been good to brush up these skills again: learn something every day!

kaste:
🥂

PR #45, opened 2025-07-04
Pretend search results are sync

### LINKED ISSUE #42: Prevent content flash on navigation back to search results
https://github.com/packagecontrol/thecrawl/issues/41#issuecomment-3034918769

Basically the page loads with the static content, and then search queries are considered and applied. Usually this is super fast, but in some cases a flash can be noticed.

kaste:
Fixes #42

If the URL does not have a "q", the page shows a statically homepage. If "q" is set, it enters into search results mode but only after all the js has loaded.  This results in content flash.

We do two things if "q" is present.

(1) we immediately hide the newest and recent section, but also the counter.  For the counter, there must be new code that generally undoes that.  This is done in `setCounter`.  We also fill in the "q" value into the search input because that looks nice.

(2) we mark the whole document as "bootstrapping", and use that to hide the "main" and the "footer" section.  We do this until the end of "index.js" but for not longer than 1 second.

Typically, the results come in fast enough and the page will just render without any layout jumping or content flashing.  If it takes more than 1 second, it will basically show pa skeleton page: the input box will have the correct value, the footer will appear... but nothing more because of the changes we did in (1).

### DIFF
--- a/index.njk
+++ b/index.njk
@@ -5,6 +5,19 @@ date: Last Modified
 
 {% import "packages/macros.njk" as pack %}
 
+<script>
+  (function(){
+    const q = new URLSearchParams(location.search).get('q');
+    if (q) {
+      document.documentElement.setAttribute('bootstrapping', '');
+      document.write('<style>section[name="newest"], section[name="recent"] { display: none; } h1 { visibility: hidden; }</style>');
+    }
+    setTimeout(() => {
+      document.documentElement.removeAttribute('bootstrapping');
+    }, 1000);
+  })();
+</script>
+
 <h1 data-all="{{ collections.packages | length }}">
   {{ collections.packages | length }}
   Packages
@@ -31,6 +44,15 @@ date: Last Modified
 
   {{ pack.labels(['linux','macos','windows'], ['language syntax','snippets','linting','auto-complete','color scheme','theme']) }}
 </form>
+
+<script>
+  (function(){
+    const q = new URLSearchParams(location.search).get('q');
+    if (q) {
+      document.forms.search.elements['q'].value = q
+    }
+  })();
+</script>
 
 <section name="newest">
   <h2>Newest</h2>
--- a/static/index.js
+++ b/static/index.js
@@ -100,3 +100,5 @@ document.addEventListener('click', (event) => {
     }
   }
 });
+
+document.documentElement.removeAttribute('bootstrapping');
--- a/static/module/list.js
+++ b/static/module/list.js
@@ -19,10 +19,10 @@ export class List {
     const counter = document.querySelector('h1');
     if (count === 1) {
       counter.innerText = '1 Package';
-      return;
+    } else {
+      counter.innerText = (count ?? counter.dataset.all) + ' Packages';
     }
-
-    counter.innerText = (count ?? counter.dataset.all) + ' Packages';
+    counter.style.visibility = 'visible';
   }
 
   // reveal search results and hide the static homepage
--- a/static/styles.css
+++ b/static/styles.css
@@ -11,6 +11,13 @@ html {
   height: 100%;
 }
 
+html[bootstrapping] header,
+html[bootstrapping] main,
+html[bootstrapping] footer {
+  display: none;
+}
+
+
 body {
   background: var(--background);
   color: var(--foreground);


kaste:
This leaves actually room for a spinner (after 1 second).  Doesn't have to be an annoying spinner, any message would suffice but that's it for now.  Holidays 🎶 I may be curious enough to look at something here but I will not have a keyboard with me.  'til August.  🌞

braver:
> 'til August. 🌞

Well deserved! See you in August 😉

braver:
Had some time to look at this, and I get what you're after. I think I'd implement it differently though (using custom events rather than hardcoded timings) ... I'll have a look.

braver:
https://github.com/user-attachments/assets/fd944061-0dbf-4e96-ae5b-e5053934e4da

braver:
Note that this is severely throttled, and with caching turned off, to make the transition clear. Normally it should not be this bad 😅

braver:
resolved via f1377c268b159ce35a965493bb8100ddb6fc174c

kaste:
Just refreshing a search page flickers with your approach.  The artificial wait time is there "obviously" to decide if we should show a spinner - in your case the "searching" - at all.  

What I was trying to do is: simulate a search page statically delivered from the server.  That is the best UI/UX in such cases as it is in fact the loading that is slow not the searching.

braver:
> Just refreshing a search page flickers with your approach.

I wouldn’t consider that an enormous problem, but I can look into perhaps optimizing further. I am wary of fighting the overall architecture too much, or spending too much time optimizing details when major functional aspects (like download/install stats, a libraries list, etc) are still missing. But, I can definitely look into tweaking some more.

braver:
To clarify the above… I’m also thinking about the request to have the search field on all pages, and am considering the idea that all of this points to integrating search on the homepage is simply thee wrong design after all. In other words, we might have to design our way out of the problem rather than bug fix and dig ourselves deeper into this hole.

braver:
It's going to blink one way or the other right? 

This PR:
https://github.com/user-attachments/assets/4b6db306-01e0-4b26-afe5-b585ac4e6b0a

Currently live:
https://github.com/user-attachments/assets/7123ce9f-f707-4edb-afc0-5afe9e34dfd7

There is a static document ready to go immediately, and some JS that necessarily runs late (because it needs to do DOM manipulation). If there is a search query, we know the static document is going to be replaced eventually (and hopefully very soon), so we hide it until the search query has been processed. Hence a blink. What we do in the mean time can be focussed on minimizing the "feeling of blink", but we can't take it away. I changed the H1 to indicate what you're waiting for in the mean time, but this might feel more like blinking rather than less. Perhaps that should be done only after some time, or a less intrusive "please wait" indicator can be used.
Or we're simply taking SPA ideas and expectations into something that is not that and we need to reconsider our approach completely. 

There are a couple more pressing matters I want to look into first before I cycle back to this one, and let it percolate in the back of my mind for a while.

kaste:
With a hot cache this PR does not blink.

braver:
> With a hot cache this PR does not blink.

You mean because everything is hidden right? The main thing to do then would be to delay putting “Searching..” in the H1.

kaste:
It is not just the H1.  It is shifting.  Currently 

https://github.com/user-attachments/assets/cf39c76e-c2d8-432d-b75f-d5248eb77b72

Slow motion

https://github.com/user-attachments/assets/289bc1ca-4e76-4edc-920c-3d94a8fe3d6a

PR 


https://github.com/user-attachments/assets/e925b8dc-3ecd-4da3-a2c4-85b939b82dc4



in slow motion



https://github.com/user-attachments/assets/61422a46-cc31-42a2-9405-ff805e66192f

braver:
Er, we can open another issue if needed, but I’m already looking into it between other projects today.

braver:
I haven't really identified what makes the page shift. It's not 100% of the time for me. It looks to me like the content replacement when the search comes in that might do it (it's already there in earlier versions). Of course, if you hide (almost) all the content until that happens, you don't see the shift (because you're making it look like it's sync and blocking the page from loading). The problem with that though is that if you navigate to a package and then back to the search results, you don't land where you're supposed to: you always end up at the top of the home page:

This PR

https://github.com/user-attachments/assets/a01305ec-6fae-4d3a-a367-5367dac14fbf


[Stuff I'm currently working on](https://github.com/packagecontrol/thecrawl/tree/ongoing-developments)

https://github.com/user-attachments/assets/edf0788b-124c-440e-8e9b-156fc1e9cb27


So, yeah, haven't found the ideal trade off yet, but I'll keep an eye on it.

kaste:
Oh, the scroll bug is shit.   But I think it works with your commit just by luck, do you know how that works?  My guess is that the search listing is shorter than our home page with the newest and recent section, and with your visibility hidden vs my display none, you reserve the space for these sections so that the page in total is always long enough for the scroll.  (?)  Usually with dynamic content we would need to save and apply the scroll position manually, no?

braver:
Yes it might just be luck! Otherwise it would be some timing thing. It’s pretty precise though for it to be pure luck 😅. It’s probably different across browsers. For instance I notice that on Safari it will render from memory and won’t even execute the scripts in the head, when you navigate back… but like half the time…
We could indeed remember and attempt to restore the scroll position, but doing so quite late after the page has loaded can get janky… maybe ensuring the page is always at least as long is the “cheaper” and more reliable way (even if it’s luck more than engineering 😉).

PR #48, opened 2025-07-06
Add package stars format

ocarreterom:
This PR formats the GitHub stars displayed for each package.

If the same format is to be used for the installation counter, perhaps the name should be changed to something more generic.

Captura de pantalla 2025-07-07 a las 1 14 27

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -147,8 +147,8 @@ module.exports = function (eleventyConfig) {
   });
 
   eleventyConfig.addFilter("stars_format", (count) => {
-    const starsFormatter = new Intl.NumberFormat("en", { notation: "compact" })
-    return starsFormatter.format(count)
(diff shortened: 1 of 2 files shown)


.eleventy.js
```diff
@@ -146,6 +147,10 @@ module.exports = function (eleventyConfig) {
     return (new Date(date)).toDateString();
   });
 
+  eleventyConfig.addFilter("stars_format", (count) => {
+    return starsFormatter.format(count)
```
braver, line 151:
```suggestion
    const starsFormatter = new Intl.NumberFormat("en", { notation: "compact" });
    return starsFormatter.format(count);
```

braver:
Thanks! Just FYI, search results are rendered in JS (`static/module/card.js`) and they need this formatter as well. I'll take care of that.

PR #49, opened 2025-07-08
Add "sort by" select outline

ocarreterom:
Add focus outline to "sort by" `<select>`.

Captura de pantalla 2025-07-08 a las 16 19 43

Captura de pantalla 2025-07-08 a las 16 20 41


braver:
Thanks! Yeah I had to really squeeze it in there, but if .8em works, then let's go with that!

PR #51, opened 2025-07-12
URL-encode syntax paths

michaelblyons:
Make all those spaces into `%20`s. Fix the hard ones, too. Some packages actually have unprintable characters in the path.


kaste:
@braver I asked Terminal to push the code so I (or obviously as it seems others) can have a look.

PR #71, opened 2025-08-05
Effectively filter all ST2-only packages

deathaxe:
This PR proposes to extend filter logic of registry generation script, to filter out all ST2-only packages or release branches.

This drops another chunk of 367 old packages.

### DIFF
--- a/scripts/generate_registry.py
+++ b/scripts/generate_registry.py
@@ -79,15 +79,13 @@ async def fetch_packages(channels: list[str], db: Registry = None) -> Registry:
         repos: list[str] = flatten(repos_lists)
         unseen = Unseen(repos)
         sem = asyncio.Semaphore(MAX_CONCURRENCY)
-        result: dict[Url, RepositorySchema] = {}
-        result = {
+        result: dict[Url, RepositorySchema] = {
             repo["self"]: repo
-            for repo in await asyncio.gather(*[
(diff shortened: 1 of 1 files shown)


kaste:
I don't think we should rewrite the registry to *drop* things because (a) we could probably just remove them at package_control_channel which would be less magical for the users, but (b) we show these ancient packages on the website for historical, nostalgic reasons (no, for the curious users that have them still installed or in a blog post, I don't know).  However, they don't appear on the compiled, shipped channel.json.  At least I hope so, random test showed that that is the case.  

The schema_version test is correct.  However, the direction I took was that the registry actually stores for each package its schema and origin (as "self"), so instead of disallowing a mismatch, we can just allow them.  They are just mis-tagged right now because I did not consider that.  (Right now, the schema_version is actually *not* used by the crawler, it is just informational.)   

Anyway, dropping must be discussed somewhere I think if it affects the website.

braver:
Guys, please keep in mind that tone is hard to judge online and for most of us English is not our first language. If something reads as an attack it very often isn’t. 

I enjoy working on this project but a requirement for my involvement is that we try to get along. Note that this is all open and public, and that open hostility is likely to scare away both potential and currently active contributors. Please try to resolve differences via email, DM each other on Discord, go grab a bear together virtually or for real, or something similar, and keep discussions in here positive and focused on moving the project forward.

The above is not aimed at anyone in particular but a general statement and request.

PR #77, opened 2025-08-11
Display archived state

### LINKED ISSUE #76: Read the "archived" flag on Github repos and show that
The crawler could read the "archived" bit, if we find a way to show that on the website and find that info useful.

braver:
fixes #76 

Tried a couple different things, I think this is a relatively unobtrusive approach that still clearly labels archived repos as "buyer beware".

Scherm­afbeelding 2025-08-11 om 13 27 41
Scherm­afbeelding 2025-08-11 om 13 30 17

### DIFF
--- a/static/module/card.js
+++ b/static/module/card.js
@@ -44,7 +44,6 @@ export class Card {
       warning.setAttribute('title', 'Package was never crawled')
       warning.querySelector('.counter').innerText = 'R.I.P.'
       this.clone.querySelector('h3').innerHTML = this.pkg.name
-      this.clone.firstElementChild.classList.add('dead-on-arrival')
     }
     else if (this.pkg.removed) {
       const date = new Date(Number(this.pkg.removed) * 1000)
@@ -114,10 +113,6 @@ export class Card {
     else {
       a.classList.add('button', 'label')
       a.setAttribute('href', '/?q=' + encodeURI('label:"' + name + '"'))
-    }
-
-    if (name == 'Archived') {
-      a.classList.add('archive')
     }
 
     a.innerText = name
--- a/static/style/buttons.css
+++ b/static/style/buttons.css
@@ -51,7 +51,3 @@ span.button {
     background: var(--background-3);
   }
 }
-
-p > .button {
-  display: inline-flex;
-}
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -64,16 +64,16 @@
     padding-right: 4.8em;
   }
 
+  h3 {
+    color: var(--foreground-3);
+  }
+
   h3 a,
   a[href*="q=author"] {
     color: var(--foreground-1);
     &:hover {
       color: var(--orange-1);
     }
-  }
-
-  &.dead-on-arrival h3 {
-    color: var(--foreground-3);
   }
 
   p:last-child {


kaste:
Holy, that looks nice.

kaste:
"Archived" could be understood as not installable though.  "abandoned" often has the negative reading ("unmaintained", "unowned" not so.)  "discontinued" is also good.

kaste:
The same style could be used for "removed" packages, these are actually "archived" in the sense that they're no installable anymore; just that nobody claimed them yet.

michaelblyons:
Do you have a list of states a package could have? I'm having trouble keeping them straight between "tombstoned," "removed," "archived," "missing," etc.

Another possibility for this one is "frozen" (or "source frozen")

braver:
> Do you have a list of states a package could have

That’s a good question actually… I’ll have a look at the different ways a package may “end”.

> "discontinued" is also good

That’s really good actually.

kaste:
There is no "missing" as we we use the GraphQL API which has better error handling.  If the query throws, we know that the API just died.  Otherwise we get a clear, valid response saying e.g. that repo has been removed.  So we literally don't get 404's for them.

michaelblyons:
Discontinued makes sense from *our* perspective. It is almost literally true.

But from a visitor's perspective, I think it gives the wrong impression. Most things that are discontinued have a dwindling stock or are unavailable, neither of which is the case here. We can keep serving the same old (or even not-so-old) package to anyone who wants it.

It's not a bad word, but I'm hoping for something better.

braver:
Yeah, it’s kind of “frozen” but I’m not sure if you see that label for the first time that it would make sense to you.

Also, Disney connotations…

braver:
“Abandoned” in the sense of abandonware is something that most devs will immediately understand. Sounds a bit like a judgement on it… like you’re leaving it out in the rain all alone 😅

kaste:
Since it says removed ... and it is RIP, we should say Created/Born Nov 2018, Found dead Dec 2024.  At least something like that. You get the point.  Not last modified at least.

kaste:
@michaelblyons Take https://packagecontrol.io/packages/One%20Dark%20Color%20Scheme This one is not even marked missing although it is a 404 *and* even removed from the registry.

kaste:
I think the whitespace is wrong -- doesn't belong here but anyway --  Between the heading and the subheading, name and details, that seems too wide.

image

That doesn't need to be correct as I don't have *any* design tools on my machine, but name without description doesn't tell anything so they maybe should be closer.  :sorry_for_the_noise:

kaste:
I still have words... they're just coming... 

I think the UI should encourage authors of packages to not delete their
work, one button click, but to mark it archived, the button above that.

And users should still be encouraged to use it, but then to fork and maintain 
it. So the word, as always, must be positive.

Then maybe rather UNMAINTAINED, not DISCONTINUED.

static/module/card.js
```diff
@@ -88,9 +116,18 @@ export class Card {
       a.setAttribute('href', '/?q=' + encodeURI('label:"' + name + '"'))
     }
 
+    if (name == 'Archived') {
+      a.classList.add('archive')
+    }
+
```
kaste, line 122: I think this is not used

kaste:
With a long enough name, the layout of the card has a bug:  

image

No idea, where that comes from.  The pic shows UNMAINTAINED, I pushed that as a commit as well.   Easy to revert but the layout bug is there anyway.

kaste:
For a tombstone, we now have the minimal

image

kaste:
I think the DOA solution is quite good.  We could also just remove these few packages.  We clearly also don't remember any other package that has been removed from the registry before our crawler was there.  (and we will never get access to the original db, if pc.io even recorded all of that, for all the time.)

Generally, LGTM.

kaste:
Well, hard to judge the css changes.  And some other bits, as this PR got bigger and also has I think unrelated changes.  Then we just need to merge rather quickly before it gets chaotic.

michaelblyons:
I am not a big fan of all-caps, but it's a minor thing to quibble about.

kaste:
Just remember the name, DOOM.   Yea.  I'm confused by the RIP, which is all caps, I think.  "Unmaintained" is a very long word here on the card.  And without caps it looks like a word not a tag.  We basically, don't want to read that "label", we "tag" the card.  I think that should be graphical somehow.  Maybe that just takes two days or weeks more.  The PR gets too long, however.

kaste:
Picture of a failing state/reason, in your face

image

These reasons could be tracebacks(! -- heck, we're open source, and if in doubt I just put the actual error plus python stack in there. Only for the known unknowns I have nice messages --) and any ASCII art and therefore should probably be formatted as "code" or "pre".  E.g.

image

michaelblyons:
RIP is an acronym for Rest In Peace (or the Latin equivalent).

braver:
I don't know about the all caps. I get the idea, but with the red it's (to me) like a shout or warning, whereas in normal type it's just a statement of fact. I don't like how small-caps look either. The word is just too large. Maybe we should call it `M.I.A.` 😅 We can bike-shed for weeks over this!

PR #79, opened 2025-08-11
Fetch and accumulate stats when crawling

kaste:
Regularly fetch the stats and compute dailies, weeklies, and so on.  Publish as "stats.json" under https://github.com/packagecontrol/thecrawl/releases/tag/crawler-status.

Also make backups of the crawler's workspace.  (However we could restore them.)


kaste:
Let's roll it.  Find it here:

image

braver:
🚀

PR #83, opened 2025-08-17
Format dates in a "humanized" style

### LINKED ISSUE #81: Humanized dates
Linus invented humanized date formats for git, I think we could apply them here.

The first line shows what we have today.  But as theses dates are from a long time ago, we can
reduce their information.  

The second block shows that in table format.  The third block shows how the humanized format
changes with the distance to that datum.  For very old dates, or in some contexts, just showing the
year might be the most elegant format even. 

```
Since January 3, 2017, last updated February 26, 2017, archived January 6, 2023.
Since January 2017, last updated Feb 2017, archived Jan 2023.
Created 2017, Removed 2024.




Created      Jan 2017
Last update  Feb 2017
Archived     Jan 2023


Last update  2 minutes ago
Last update  yesterday
Last update  Aug 12, 2025
Last update  Aug 2024
Last update  2016
```

kaste:
Fixes #81

Examples:
```
Last update  2 minutes ago
Last update  yesterday
Last update  Aug 12, 2025
Last update  Aug 2024
Last update  2016
```

image

Aside:  Why is the version stuff not tabular.  E.g. `1.2.1 (ST4)    Aug 2024`


michaelblyons (changes requested):
Tiny whitespace regression.

michaelblyons:
> Aside: Why is the version stuff not tabular. E.g. `1.2.1 (ST4)    Aug 2024`

I like this in general, but for a single version it might look strange to have a heading and only one row beneath. We might also want to rearrange to put ST version first.

| ST versions | Package Version | Release Date |
|-|-|-|
| >=4075 | [6.2.0](https://codeload.github.com/robballou/sublimetext-sshconfig/zip/st4-6.2.0) | yesterday |
| 3156 - 4074 | [4.1.1](https://codeload.github.com/robballou/sublimetext-sshconfig/zip/st3-4.1.1) | 2020 |
| 3000 - 3155 | [2.0.1](https://codeload.github.com/robballou/sublimetext-sshconfig/zip/st3.0-2.0.1) | 2019 |
| <3000 | [1.1.3](https://codeload.github.com/robballou/sublimetext-sshconfig/zip/st2-1.1.3) | 2015 |

If we go tablular, do we still need the big **Versions** heading, or just move the table above **Links**?

braver:
> Why is the version stuff not tabular

Because most times there’s only one version, and it’s not like it’s an enormous spreadsheet. But, primarily because tables are hell on smaller screens and hell for accessibility. I avoid them unless I really can’t.

braver:
Overall, while this is neat, I have the issues with it that I mentioned in #81. I also don’t like the abbreviated months. “Created July 30, 2020, last updated July 9, 2025” reads to me like an entirely normal sentence that I can simply understand easily. If I don’t care when exactly in 2020 I can ignore it, and if I do care the information is not hidden from me.

Maybe this is unreasonable or inflexible from me, but it’s removing information from the UI for no apparent benefit. As a UX designer I just don’t entirely get the goals of this change.

braver:
I do like the relative time (minutes, hours) and the “yesterday”. 
My issue is mostly about hiding dates that are further away. When I’m reviewing packages these are regularly quite relevant. Similarly, the dates on versions include the time on purpose, because again that detail is sometimes relevant. And they’re not in a language-y context, so here a developer can easily and unambiguously verify the precise timestamps across several releases.

In any case, if any of that detail is hidden, I’d like to see an (obvious) way to reveal it, even on a mobile device.

braver:
Sorry! I sound like an awful spoilsport 😓 I just really regularly use that data 😅

michaelblyons:
> This went completely off the rails.

Sorry about that!

> the liftetime dates section

Good point, especially with the three-date sentence. How about up there it's a date like your algorithm, maybe with the stippled underline like FF mentioned, and if you click it expands to something longer? (How much longer it should be I leave to Braver.)

>  combine this and drop the versions section for the 80% case.

I have mixed feelings. I do think that's the most common, but I think it would also be good to show the ST versions, at least if requested.

FF, do you have a sketch of what you mean with the "effective" version range? The things that I imagine are all misleading before you do the expansion. (Basically they suggest that the whole range is updated at the newest version's update time. I don't know about you, but I tend to leave old ST versions behind when I start a new tag series.)

> "last updated" could also mean -- since thecrawl is much more capable -- most recent commit date on the default branch. Then it could highlight, hey the authors never cut a new release. I should check their main branch.

This is interesting. I think worth trying and low impact if you have to stop.

> I would prefer when the version would not link to the zip but to the actual tag/release on GitHub. 

Agreed. Can you tell if a tag is a release to pick which place to link? I never make formal releases (just tag the versions), but I know some people have releases with changelogs and assets.

braver:
I like the more readable "Created ... " sentence. I just don't want to completely remove the exact date (so you can celebrate a package birth day :) ), or times on versions. If that information is completely removed it's not recoverable for whatever emergent use case someone might have. All my complaints are gone when the exact date can be revealed as FichteFoll described (stippled underlines or something else).

braver:
> since thecrawl is much more capable -- most recent commit date on the default branch. Then it could highlight, hey the authors never cut a new release. I should check their main branch.

Oh that makes sense. The last updated now is not the last commit, it’s the last tag? I didn’t even realize that.

braver:
> I would prefer when the version would not link to the zip -- I never used that, if in doubt I clone with git -- but to the actual tag/release on GitHub

Makes sense! I just linked to the zip because I could, but what you describe is indeed much more useful.

michaelblyons:
> Then you would show something like "Sublime Text 4070+" if the package had releases for `4070 - 4199` and `>=4200`. The fact that the release for an older ST might not have been updated as well is negligible in the overview and still accessible when expanding the detailed version breakdown.

So for "Sublime Text 4070+" you wouldn't show any version number until the expansion? That's not something I had thought about. :slightly_smiling_face: I guess it makes sense that people do not care what your version number is most of the time.

The current ReadMe might not be relevant, but that's a problem that previously existed.

packages/package.njk
```diff
@@ -39,23 +39,24 @@ eleventyComputed:
...
-      , archived {{ pkg.archived_at | date_format }}
+      , archived <human-date datetime="{{ pkg.archived_at }}">{{ pkg.archived_at | date_format }}</human-date>
     {%- endif %}.
 
   {% elif pkg.last_modified %}
     <br>
-    Last updated {{ pkg.last_modified | date_format }}{%- if pkg.archived_at -%}
-      , archived {{ pkg.archived_at | date_format }}
+    Last updated <human-date datetime="{{ pkg.last_modified }}">{{ pkg.last_modified | date_format }}</human-date>{%- if pkg.archived_at -%}
+      , archived <human-date datetime="{{ pkg.archived_at }}">{{ pkg.archived_at | date_format }}</human-date>
```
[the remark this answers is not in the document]
kaste: Forgot about that yesterday.  So, we had leading commas before and I also don't see any additional whitespace on the rendered page.  Do you?  Or is that a source code formatting issue?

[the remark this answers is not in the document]
michaelblyons: If your screenshot is from this code, these are not needed. I would have expected that this HTML would result in a space between the date and the comma, but maybe Nunjucks removes leading spaces.

_includes/scripts.njk
```diff
@@ -40,3 +40,229 @@
...
+        const abbreviate = this.hasAttribute('abbreviate-months');
+        const monthFormat = abbreviate ? { month: 'short' } : { month: 'long' };
+        const monthFmt = new Intl.DateTimeFormat('en-US', monthFormat);
+        const yearNum = new Intl.DateTimeFormat('en-US', { year: 'numeric' });
+        const hm = new Intl.DateTimeFormat('en-US', { hour: '2-digit', minute: '2-digit', hour12: false });
+
+        if (!alwaysMonths && ms > 7 * year) {
+          return yearNum.format(dt);
+        }
+        if (ms > 1 * year) {
```
braver, line 188: Just cause I'm curious, is this logic based on a reference implementation?

kaste: It is simplified a lot.  Typically, after yesterday, you'd switch to "Mon 12:11" for 5 or so days.  Then "April 1, 2025" if the year is this year...  The last switch to "2017" is also arbitrary late because we're usually have not that many updates and I didn't want to switch too early (=too often).

_includes/scripts.njk
```diff
@@ -40,3 +40,229 @@
...
+        const d = new Date(normalized);
+        if (!isNaN(d.getTime())) {
+          return d;
+        }
+        const d2 = new Date(raw);
+        return isNaN(d2.getTime()) ? null : d2;
+      }
+
+      _format(now, dt) {
+        const ms = now - dt;
```
braver, line 166: It looks like you're missing timezone handling. Dates in the data are UTC. It's now 17:48 in UTC, 19:48 locally. It should say "just now", but it says "2 hours ago". If I put my local current time in the `created_at` (which would be 2 hours from now) it does say "just now".

Scherm­afbeelding 2025-08-21 om 19 47 48

I wonder, if we're now doing all date presentation stuff client-side anyway... can't we just present everything in local time? (And change the comment about that in the footer as well)

_includes/scripts.njk
```diff
@@ -40,3 +40,229 @@
...
+          }
+        }
+      };
+    }
+
+    class HumanDateElement extends HTMLElement {
+      static attrs = {
+        "abbreviate-months": "bool",
+        "always-months": "bool",
+        clickable: "bool",
```
braver, line 100: Currently the "reveal raw date" feature is not accessible via keyboard. Can we make these elements tab-focusable and respond to return/space keypresses? That means it should also probably have a button role and an indication of its purpose ("click to toggle between raw and readable date").

https://www.w3.org/WAI/ARIA/apg/patterns/button/

_includes/scripts.njk
kaste: Local iso-ish formats can be confusing:

image

However, the humanized format is local.  It would show "today, 17:29".

braver: > now it is anything or nothing

As far as I can tell, if you’re running in GitHub actions it’s UTC. The build timestamp of the site definitely is. It must be some kind of timezone deterministically, otherwise we should not be mentioning times and even dates become questionable. Or I don’t understand your point.

braver: Oh, you’re saying they’re UTC “by convention” because we know they are, but since the timezone is stripped they’re not UTC explicitly? I must say I find that really inelegant about our spec… what use is a timestamp if you don’t know what timezone it’s in? Maybe we should make a v5 schema 😉

kaste:
We now have options, as we're getting timezone aware, for the *long* format.  

1. We use an iso-ish format with local timezone.  That can lead to 

image
image
image

2. The long format becomes (GitHub style) something like "Aug 21, 2025, 7:43 AM GMT+2".  Like the human format, just in explicit, long form. Always expanded you could say.

image


3. We switch to explicit UTC

image
image

This also affects our footer.  E.g 

image

 If we want to use two variants, I just add another argument to the component.

kaste:
The q is maybe, what is the design choice for these iso formats right now?  Obviously switching these two formats

image 
image

feels more "calm" as it basically just expands.

_includes/scripts.njk
braver: Thanks, that’s awesome. I’m going to run over this with my screenreader etc. It’s my new hobby, listening to what websites sound like 😄

braver:
> feels more "calm" as it basically just expands.

Yeah that looks pretty good, it’s not disorienting like switching to 8601 timestamp would be. If you’re a robot and want raw data, go read the channel itself.

You should publish this as an npm module, it’s so flexible already 😉

static/styles.css
```diff
@@ -159,6 +159,11 @@ pre {
   color: var(--foreground-3);
 }
 
+human-date[clickable] {
+  border-bottom: 1px dotted #ababab;
```
braver, line 163: Try this, it will introduce a new 4th grey we can use here:

```diff
diff --git a/static/style/colors.css b/static/style/colors.css
index 5b884dae..40a38a96 100644
--- a/static/style/colors.css
+++ b/static/style/colors.css
@@ -15,7 +15,8 @@
 
   --foreground-1: hsl(240 5 29); /* black */
   --foreground-2: hsl(240 4 40); /* ..    */
-  --foreground-3: hsl(240 3 51); /* grey  */
+  --foreground-3: hsl(240 3 51); /* ..    */
+  --foreground-4: hsl(240 3 60); /* grey  */
 
   --white: #fff;
 
@@ -41,6 +42,7 @@
   --foreground-1: hsl(60 10 84);
   --foreground-2: hsl(60 5 71);
   --foreground-3: hsl(60 3 59);
+  --foreground-4: hsl(60 2 48);
 
   --red-1: hsl(7 92 54);
   --red-2: hsl(7 98 56);
diff --git a/static/styles.css b/static/styles.css
index 728f3b6b..325301f1 100644
--- a/static/styles.css
+++ b/static/styles.css
@@ -160,7 +160,7 @@ pre {
 }
 
 human-date[clickable] {
-  border-bottom: 1px dotted #ababab;
+  border-bottom: 1px dotted var(--foreground-4);
   cursor: pointer;
 }
```

_includes/scripts.njk
```diff
@@ -40,3 +40,273 @@
...
+          timeZoneName: "short"
+        });
+        const github_style = longFormat.format(dt).replace(" at", ",");
+        const showRaw = this.hasAttribute('raw');
+        if (showRaw) {
+          this.shadowRoot.textContent = github_style;
+          this.title = formatted_text;
+        } else {
+          this.shadowRoot.textContent = formatted_text;
+          this.title = github_style;
```
braver, line 283:
```suggestion
          this.title = formatted_text + ' (click to reveal)';
        } else {
          this.shadowRoot.textContent = formatted_text;
          this.title = github_style + ' (click to reveal)';
```

_includes/scripts.njk
```diff
@@ -40,3 +40,273 @@
...
+      constructor() {
+        super();
+        this.attachShadow({ mode: 'open' });
+        this._timer = null;
+      }
+
+      connectedCallback() {
+        this.setAttribute("role", "button");
+        this.setAttribute("tabindex", "0");
+        this.setAttribute("aria-label", "Toggle between relative and absolute date")
```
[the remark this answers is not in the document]
kaste: That is correct.  Example `<button aria-label="close window">X</button>` (or delete etc.)  Because the content is *not* sufficient for loud reading.  

Actually, what happens when we do not set the role?  Anyway, we're after aria-describedby I think.  Canonical:

```
<p>
  This was last updated
  <button
    type="button"
    aria-describedby="fmt-hint"
  >
    Aug 1, 2024
  </button>.
</p>

<span id="fmt-hint" class=".hidden">Toggle date format</span>
```
Is your screen reader always reading the title?  "last update April 1, 2024 *button*"  I'm not sure why it should read the title?  Actually, I was fooled by https://www.google.com/search?q=The+title+attribute+is+not+consistently+read+by+screen+readers+and+should+not+be+relied+upon+for+providing+accessible+names+or+descriptions.&ie=UTF-8 then.  I explicitly was told to not use "title". And 

> Rule 4: Avoid Browser Fallback
> When authors do not specify an accessible name using an element or attribute that is intended for naming, browsers attempt to help assistive technology users by resorting to fallback methods for generating a name. For example, the HTML title and placeholder attributes are used as last resort sources of content for accessible names. Because the purpose of these attributes is not naming, their content typically yields low quality accessible names that are not effective.

buttons have an accessible name, why should it fallback to title?  Anyway, I think describedby is semantically correct.

_includes/scripts.njk
```diff
@@ -40,3 +40,229 @@
 
   })();
 </script>
+
+<script>
+  (function(){
```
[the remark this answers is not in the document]
kaste: Done

static/styles.css
kaste: done

packages/package.njk
```diff
@@ -39,23 +39,24 @@ eleventyComputed:
...
     {% else %}
-      Removed {{ pkg.removed | date_format }}.
+      Removed <human-date clickable datetime="{{ pkg.removed }}">{{ pkg.removed | date_format }}</human-date>.
     {%- endif %}
 
   {% elif pkg.created_at %}
     <br>
-    Created {{ pkg.created_at | date_format }}{% if (pkg.created_at | date_format) != (pkg.last_modified | date_format) -%}
-      , last updated {{ pkg.last_modified | date_format }}
+    Created <human-date clickable datetime="{{ pkg.created_at }}">{{ pkg.created_at | date_format }}</human-date>{% if pkg.created_at != pkg.last_modified -%}
```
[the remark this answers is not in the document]
kaste: done

_includes/scripts.njk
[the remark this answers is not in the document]
braver: I didn’t tweak anything, it’s pretty verbose by default. From what I understood from having talked to some people who use this stuff for real, is that most will tone down the verbosity quite a bit. It’s going to be different for everyone depending on what tool they’re using (I’m using Apple VoiceOver) and how they configured it. For me it did also read "last updated April 4 2024 Toggle date format Button".

braver (approved):
🎉

braver:
We did it! 😂💪🏻

PR #84, opened 2025-08-21
Change "orange-2" for better contrast/readability

kaste:
I would appreciate if the orange would be changed rather sooner than later because the contrast is too low for my eyes.  

The exact color I propose can be seen as the same color, really.  

But the contrast is already so much better, that we could switch to a lighter, or even white, background.  However, I don't remember why we have a grey, a relatively dark grey actually, as the background.  My guess is because of contrast issues.  🤷 

E.g. PR -> before

image

image

### DIFF
--- a/static/style/colors.css
+++ b/static/style/colors.css
@@ -22,7 +22,7 @@
   --red-1: hsl(7 98 56);
   --red-2: hsl(7 92 54);
   --orange-1: hsl(32 98 56);
-  --orange-2: hsl(32 92 54);
+  --orange-2: hsl(33.65deg 100% 47%);
   --blue-1: hsl(198 97 55);
   --blue-2: hsl(198 89 53);
   --green-1: hsl(80 70 51);


kaste:
If we don't aim at #fff for the background, I think `background: linear-gradient(180deg, black -100%, hsl(0 0 97) 5%);` is probably enough.  And for the cards we have now 94, `hsl(0 0 95)` is already enough, if you want dark(er) cards.

braver:
Really the colors should be ok for AA WCAG and they're not. Our contrast is already a ton better than the [docs site](https://docs.sublimetext.io) which in light mode is pretty much unreadable. Which is also where our grey background comes from. The different between white and this grey isn't actually a big deal in terms of contrast, white won't make a light color like this orange suddenly work for everyone.

Your tweak is not enough to reach AA and will not be enough of an improvement for most people. I'm ok with merging this if it helps you. But I'll make a priority of finding a color solution that's accessible for everyone.

braver:
I would go with something like `hsl(30 77 36)` perhaps:

Scherm­afbeelding 2025-08-21 om 20 17 43

braver:
Try this:

```diff
diff --git a/static/style/colors.css b/static/style/colors.css
index 5b884dae..87efb1cb 100644
--- a/static/style/colors.css
+++ b/static/style/colors.css
@@ -21,12 +21,12 @@
 
   --red-1: hsl(7 98 56);
   --red-2: hsl(7 92 54);
-  --orange-1: hsl(32 98 56);
-  --orange-2: hsl(32 92 54);
-  --blue-1: hsl(198 97 55);
-  --blue-2: hsl(198 89 53);
-  --green-1: hsl(80 70 51);
-  --green-2: hsl(80 67 49);
+  --orange-1: hsl(30 81 48);
+  --orange-2: hsl(30 78 44);
+  --blue-1: hsl(198 66 49);
+  --blue-2: hsl(198 63 45);
+  --green-1: hsl(81 47 43);
+  --green-2: hsl(81 43 39);
 }
 
 :root[data-theme='dark'] {
diff --git a/static/style/pagination.css b/static/style/pagination.css
index 8680c4d9..738f2680 100644
--- a/static/style/pagination.css
+++ b/static/style/pagination.css
@@ -19,7 +19,7 @@
   /* disabled because it's the current page */
   &:disabled {
     background: var(--orange-1);
-    color: var(--foreground-1);
+    color: var(--white);
     border-color: var(--orange-1);
     cursor: revert;
   }

```

kaste:
I had colors like hsl(30 77 36), but ...

image
image
image

That's a bit rude because it is lush not bright, and has nothing in common with orange, esp. the Sublime Text orange.  Bright orange in a light context is ... just difficult.   AA is just not possible with orange.

image
image
is hsl(19, 94%, 57%).  Actually, too red, a bit better contrast than the PR.

braver:
> I would just go with this minimal thing

Agreed. 

Links actually need to be underlined too. We can’t have our brand look and WCAG AA at the same time, or at least not without more thinking/designing. So let’s go with the minimal thing that adds some contrast, I’ll get to the other stuff at some later point.

braver:
I find the green too light as well actually. Final offer:

```diff
diff --git a/static/style/colors.css b/static/style/colors.css
index 5b884dae..1005eb99 100644
--- a/static/style/colors.css
+++ b/static/style/colors.css
@@ -21,12 +21,12 @@
 
   --red-1: hsl(7 98 56);
   --red-2: hsl(7 92 54);
-  --orange-1: hsl(32 98 56);
-  --orange-2: hsl(32 92 54);
-  --blue-1: hsl(198 97 55);
-  --blue-2: hsl(198 89 53);
-  --green-1: hsl(80 70 51);
-  --green-2: hsl(80 67 49);
+  --orange-1: hsl(32 94 55);
+  --orange-2: hsl(32 92 49);
+  --blue-1: hsl(198 97 54);
+  --blue-2: hsl(198 88 51);
+  --green-1: hsl(80 70 49);
+  --green-2: hsl(80 67 47);
 }
 
 :root[data-theme='dark'] {
 ```
 
Enough to matter, too little for most people to notice. For the links it comes to a 2.29:1 contrast ratio, slightly better than your proposal (2.28).

Before

Scherm­afbeelding 2025-08-22 om 21 17 03

After

Scherm­afbeelding 2025-08-22 om 21 21 58

kaste:
Haha.  Subtle.  A nerd thing, I guess.  But indeed it is orange, basically the same color.  🚢 it.

What's the story behind the grey background.  What problem does that solve?  I don't remember. Is it also a trademark, signature color from the main page?  Because of the cards?  (Well, it is basically a signature color now, nobody has that.)

Issue #34, opened 2025-07-01
Make search field available on all pages


kaste:
Usually, search is at the top always available, and I also use it like that.  Why should I click on the top logo, goto the main page, just to search again. 

I see ctrl+k is often bind to enter the search.  If so, make it so.

kaste:
Oh.  Doesn't even work, `s` for search works.

braver:
Shipped via #93

PR #91, opened 2025-08-28
immediately go back to search and start typing

braver:
An alternative approach to #89 and #34.

The interaction is similar: click search ➡️ start typing.

Because of caching, there is very little delay between the click and you can start typing. Of course there is the context switch making it slightly less smooth. But you get a lot in return: 

- search behaves the same way wherever you search (because it _is_ the same)
- you always get the sorting field
- you always get the labels to click
- we retain the H1 and results counter
- the urls and browser history makes sense (search results are in one place, detail pages are elsewhere)
- it also works on smaller devices
- you can continue your previous search action and refine if needed, or start over

And we get in return that it takes hardly any code. Actually it's just the `#search-field" anchor in the link that makes it work, the rest is just flourish.

https://github.com/user-attachments/assets/581ac975-4e17-4b43-82ad-cf9af36bc48f

### DIFF
--- a/_includes/header.njk
+++ b/_includes/header.njk
@@ -2,7 +2,7 @@
   <div class="inner">
     <nav>
       {% if page.url != "/" %}
-        <a href="/">Home</a>
+        <a href="/#search-field">Search</a>
       {% endif %}
       <a href="/about">About / FAQ</a>
     </nav>
--- a/packages/package.njk
+++ b/packages/package.njk
@@ -108,5 +108,6 @@ eleventyComputed:
   {% if pkg.readme %}
   <section id="md" class="readme" data-readme-url="{{ pkg.readme }}"></section>
   <script src="{{ '/static/readme.js' | bust }}" type="module" async></script>
+  <script src="{{ '/static/tosearch.js' | bust }}" type="module" async></script>
   {% endif %}
 {% endif %}
--- a/static/styles.css
+++ b/static/styles.css
@@ -10,6 +10,7 @@
 
 html {
   height: 100%;
+  scroll-padding-top: 65px;
 
   --side-padding: 1rem;
   @media (min-width: 78rem) {
--- a/static/tosearch.js
+++ b/static/tosearch.js
@@ -1,0 +1,6 @@
+const link = document.querySelector('[href="/#search-field"]')
+const referrer = new URL(document.referrer)
+
+if (link && referrer.host === window.location.host && referrer.search) {
+  link.href = '/' + referrer.search + '#search-field'
+}


kaste:
Is this for real?  😂 Every site has a search field in the header.  Do you set the bar to high for the first iteration?  Obviously I did notice the navigation hiccup, but I did not push any commits to thecrawl since Monday anyway. 

> search behaves the same way wherever you search (because it is the same)

The search itself is the same.  I'll post two result views so I can look at them actually.  Didn't had a look since the first push.

image
image

Well, that's (a) pretty close already, and (b) looking good.  ( That looks actually better than what I usually do in one hour 🤣 )

>you always get the sorting field

Yeah, the typical UX is that the search grows and shrinks on focus these days.  For normal screens there seems to be enough space to grow the search input.  Easily the Sort by Magic can appear here.  The two views above already are close, but then with the full search input...  we can of course move the search field to the original position.  In the end, the search is the only SPA page, I would have moved it to a URL  "/search/?q=cool".  And both, the homepage and the details pages, then go to that site/look on enter.

>you always get the labels to click

I read them as shortcut links on the homepage.  If the links were important we could have had them on the right side e.g. under about.  I really think and thought this is content for the landing page.  In the details page, I'm used to click on "what else has this author", and then "what else has these labels" (on the details page we already show labels!).   

>we retain the H1 and results counter

the results counter is there in an h2.  It could also say "12 packages found",  "13 matching packages", ...

>the urls and browser history makes sense (search results are in one place, detail pages are elsewhere)

the nav bug I mentioned. It should be a separate endpoint "/search/?q=" I think.  

>it also works on smaller devices

Not sure what you mean

image

>you can continue your previous search action and refine if needed, or start over 

Everyone(?) goes back to get back to the search results.   I always found that strange on the old site.   E.g 
image
If I want a finer search, I need to see the search results (hence go back) and find a better term.  So obviously you *can* have the search term go with you and I thought ... nah.

braver:
> Is this for real? 😂 

Well, I wanted to consider this at least. This is low hanging fruit and does make it easier to go back and search again or refine the search, which is a feature request I really understand. I don't understand why you seem to think it's ridiculous.

There are several things I don't like about the current state of your PR. It needs a lot more design and UI code I don't have time for right now. I don't like how it changes the layout and the design of the header, the fact that the sort field is gone. I do like the idea of a search field in the header, but I'd want to do it properly and have the time for doing that. I get that we're iterating, that's fine.

So I'm ok with closing this PR if you don't think it's a good first move. Just had an idea that I wanted to put out there. .. and kind of hoped to not get laughed at for doing so...

kaste:
>  I don't understand why you seem to think it's ridiculous.

Basically, from the timing, it is a response to the PR I made, which actually has a search box in the header (plus some quirks).  If you would have done this a month ago, I would have said, thank you.  Some functionality.  Also:  if we actually already had the endpoint "/search?...".  But in this form , we actually have two links to the landing page.    

I do hate the smiley 😂 I put in there as that comes usually hard (although it is called "joy"), but at the time I *really* was not sure if that is like an April fools, essentially because I wasn't moving *my* PR and after a week sometimes people make such a joke: If we can't move our PR/feature, let's just take this.

braver:
Sure, thanks, all good :) Look, after a day of looking at this and thinking about it, even I don't like it anymore. It's too "clever" but still super janky. It's just not _good_ and it will never be. Although I do like having figured out that you can link to `#search-field` and it will actually focus the field you linked to. But that doesn't make this a great feature. 

Let's pretend this one didn't happen and see where we can take yours. I am planning some time AFK over the next week, or at least going to really make an effort to not also code every evening. But I can probably contribute a bit there... that one at least has potential :)

PR #93, opened 2025-09-01
Search everywhere iteration

braver:
This isn't done yet. For instance, the input field now goes away abruptly (just needs an animation to be added) and sometimes expectedly (like when you click "sort by"). That's not the point, I'll get back to that. And code needs cleaning up. What this PR does is bring the design changes back to "just" adding a search field to the header. It ensures there is always an H1, then an H2 (and no H2 in the header anymore), and ensures they're styled consistently. In other words, more like it was. Hope you don't see that as a step back, but there was too much going on in that top left corner and I wanted to find a way to have a meaningful H1-H2-H3 sequence at all times.
The changes in packages/package.njk read a bit as a big change, but I just made sure we were not nesting sections: the search will hide all sections that are not the `[name="result"]` section, so we don't need to wrap the readme section in another section. And it pulls the H1 out of the first section, so we can use it for the search results counter.

So, if we can agree on this as a design direction, I don't mind fiddling with the details over the coming week or so to make all the UX and all the code really nice.

https://github.com/user-attachments/assets/750f7cd5-050d-403e-8165-b2fbce9733b4


static/index.js
```diff
@@ -93,11 +93,30 @@ input.form.onsubmit = (event) => {
...
   }, 300) // .3 seconds
 })
 
+input.addEventListener('blur', () => {
+  if (input.value.length < 1) {
+    input.form.classList.remove('overlay')
+  }
+})
+
+document.querySelector('header [href="/#search-field"]').onclick = (event) => {
```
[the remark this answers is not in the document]
kaste: Maybe also stick to the style: "addEventListener"

kaste:
Since the style is very similar to what we have, and also to what my PR shows, 
this PR is maybe about the overall DOM structure.

On a general perspective, you should make the structure of the page/HTML so,
that we can easily target the dynamic content elements, and easily swap parts
of the document.  

From that, (a) we should not e.g. `querySelector("h1")`.  `h1` is generic and we 
should not rely on something being the first `h1` esp. if the context is 
`document` (E.g. `document.querySelector...`).  If the context is already a 
specific sub-DOM, e.g. a card or a section, we can of course target its first h1. 


(b) we should have sub-DOM's, containers, so we can swap package-details with
search-results with the landing-page.  That implicates that for example the
first `h1` in search-results holds the counter, the `h1` in package-details
holds the package name, and the landing page holds whatever it wants, e.g. the
information "5104 packages in the registry."  (which you shortcut to "5104
packages").

The search result counter should likely have an id, e.g. `getElementById('search-results-counter')` 
so we can find it easily in the code.

So we basically could have `#main-content` (the original static content) and 
`#search-results-content`.  And we just swap them.  Typically, these are
div's as long as they don't have any semantics but just are technical containers
to actually swap contents.  In other words: only one of them is visible at any 
time. 

After that, `#search-results-content h1` could target the results counter element.
Obviously, that is coupling function and form, so I still prefer `#search-results-counter`.


I don't think the "design"/"style" is a step back.  It is what we have, preserved.

So what follows is my reading:  While working on the feature I actually noticed that 

```
97 packages
Results
```

is not elegant and condensed enough.  Neither are the `5104 packages` on the landing 
page explicit enough.

So, now I cannot unsee it actually.  (a) the size of the registry is one information
that could be on the landing page, or always present actually, or never.  

`97 packages\nResults` is verbose and not explicit at the same time.  It is more
whitespace than info.  The information is "95 packages match the search", 
"12 results for that term", etc.  That could (b) *really* be one `<p>`, `<h1>`, `<hn>`.
The discussion about the `h`s distracts from that.  Often this information is even
omitted.  

It is like you have a structured document in mind.

```
- 5105 packages (in the registry)   (h1)
  - Newest                          (h2)
  - Most popular                    (..)  



- 95 packages (selected from all)   (h1) 
  - Results                         (h2)
```

The former makes some sense, the latter less IMO, and in general it is very firm
(or rigid?).  Maybe you meant, `95 packages (selected)\nPage 1`?  But for 
`5105 packages\nNewest...` we don't have that relationship either, we just show
two or in the future more "random" selections, never the whole thing.  (In contrast,
we actually *could* show *all* packages paginated sorted by "last_modified", then
we actually had the same structure.)

Maybe "Search 5104 packages" as the placeholder text on the landing page is a better 
fit?  Maybe that is what you want to tell us?  Whereby my sentence is - in the header 
or footer or not at all - just "5104 registered."

Then you could also say, when "theme" is already selected, "Search 267 packages" 
(ideally "Search n themes").

kaste:
Just from quick testing, it is absolutely superb.  The header in mobile 👍 

The search by button outline/border has an animation bug.  If I tab to it, the border first gets black and then fades to orange.  🤷 I prefer faster animations, but you had them in grand'ma mode for the card:hover effect already.   

The right side ("About") shifts 2 pixels down when switching from home to package.  And in mobile the other way around or so.  

...

kaste:
Instead of ".has-input" -- sorry if I don't get it -- would `input:placeholder-shown` sufficient?  Or the obscure `has(input[value=""])`.  Likely in combination with form, e.g. `form:has...`...  I can't write that down from head.

kaste:
> :hover

I had the :hover only for when you double-click (!) the Search by label.  Because a double-click selects a word and that steals -- of course -- the focus.  :fire:

Without that :focus-within and :has(label:active) was enough.  

Another tip:  Instead of input|select:blur, try form:focusout.

kaste:
> display:none;

I had this first.  But tabbing from right and tabbing from left should be the same IMO.  Well, I tried both and felt tabbing from right, from the "About" should first go to the button.  That felt better.

braver:
Yeah, I'm not doing this well organized. Cause this is not work but hobby 😉 Don't worry, any conflicts with other work are easy to resolve so far, let that be my problem. I already merged gh-pages into this locally and will indeed take this to replace your PR eventually. I understand this introduces potentially breaking changes so I'll try to get them into gh-pages before everything else, but it's not such a major change so far.

static/index.js
```diff
@@ -139,23 +139,43 @@ window.addEventListener('popstate', (event) => {
...
+    // the shortcuts in the form should provide an additional narrowing of search,
+    // not replace it
+    if (oldQuery.includes('label:') && newQuery.includes('label:')) {
+      newQuery = oldQuery.replace(/label:("[^"]+"|\S+)/, newQuery)
+    } else if (oldQuery.includes('platform:') && newQuery.includes('platform:')) {
+      newQuery = oldQuery.replace(/platform:("[^"]+"|\S+)/, newQuery)
+    } else if (oldQuery.includes('author:') && newQuery.includes('author:')) {
+      newQuery = oldQuery.replace(/author:("[^"]+"|\S+)/, newQuery)
+    } else {
+      newQuery = oldQuery + ' ' + newQuery
```
michaelblyons, line 168: Trim `newQuery` assignment here or line 172? Or do we not care about an extra space getting into the query?

kaste: That is significant. There is a leading space and a redirect then without that space which corrupts the history.  (Just click on the same label over and over.)

braver: fixed in 38fb6d4dd73abc0097adeef5cd6ca34b08e0c868

braver:
> Well, I tried both and felt tabbing from right, from the "About" should first go to the button. That felt better.

I can see that point of view, but your focus landing on an element you couldn't see before is perhaps not ideal? I don't have a strong opinion here, but I went with if you tab into the search from either left or right you always land on the search field first. Also because (currently) interacting with the sort field doesn't actually do anything: you need to query first.

So, other than that, I think I'm kind of "done" with this iteration. 

- the config file changes have already been cherry picked 
- the "label" event delegation is only active on the homepage .. which is technically not necessary, but does feel like it provides the most logical interaction where most search happens on the homepage (but you _can_ search everywhere if you want)
- handled animation issues
- I think it looks good on all breakpoints, positioning of stuff in the header is stable across pages, etc.
- I'm relatively happy with separation of concerns between JS/CSS
- binding of behavior mostly avoids query-selectors-everywhere-syndrome and is quite explicit and hopefully readable

Unless anyone else sees issues they want to have addressed I'm ready to "beta test this on the users" 🙂

kaste:
image
image

Package details > search ac > click on first result (showing package details again) > go back.

In this state

image

braver:
> Package details > search ac > click on first result (showing package details again) > go back.

I don't get that problem. Also tried throttling to check if it's a race condition. But I also don't understand how `form.elements['sort']` can not exist at that point, while `form.elements['sort-field']` does: it's the same field. Did you have some HTML in browser cache (from switching between different branches perhaps?).

kaste:
> Package details > search ac > click on first result (showing package details again) > go back.

Can't reproduce anymore.  I actually re-started the page before posting but incognito mode works, and then after manually wiping the cache it works in normal mode as well.  (Actually, a bit scary to wipe the cache.  We don't even use that much features, no cookies, no service worker...)

braver:
🤘🏻

Issue #88, opened 2025-08-23
Provide links to the tags/release pages


kaste:
Ideally, we can provide links to the actual tagged versions.  From there users can usually interact with the UI.  In GitHub land that would be: diff with previous version, download the zip.  Eventually read the release notes.  (on my own packages, GitHub would offer to edit and create release notes even.)

I prefer this over the direct download links.  (These download links point to the same content as on the "tags" page!)  

*But* for proprietary packages, we can't provide such an URL by definition, and in fact these packages don't have any other page to actually download the zips.  So we need what we have, the current link targets, for these packages.  If that's confusing, we might give a hint... using an icon...?

kaste:
Links/strings in the workspace.json.  Can't be constructed in the presentation layer with the given data (IIRC).

kaste:
Maybe:  For the GitHub, 90% case, omit the download link --  Who does that anyway? -- and have the clean look like now, and for the proprietary packages show the version plus the download icon.   Since, I generally like the orange on the version strings, maybe the version plus the icon can be packed visually (e.g. they both react on hover) to make it clear that this link is different from the 90% case link.

braver:
> Who does that anyway?

I do. It’s the quickest way of getting the packaged version as it would get delivered. I mean, I don’t use it every week, but I did last week to help me review something and it was very useful. It doesn’t need to be a major feature, we can hide it a bit.

kaste:
Ok.  I do too.  But from the Github page.  We didn't had this feature on the old site.  

We should note that -- if it is a hub zip! -- it actually is **not** what gets installed, a user needs to open the zip as it contains everything in a sub-folder, and extract that sub-folder to the correct place,  renaming the folder on the go, t.i. stripping the version.  

This is in stark contrast to the proprietary download links, e.g. https://packages.sublimetext.io/packages/Theme%20-%20Monokai%20Pro/ for which we provide a ready to use *.sublime-package file, e.g. Theme%20-%20Monokai%20Pro.sublime-package

!

(It would be interesting if we *could* deliver ready-to-use package files.  Actually, initially, I thought the crawler would do the repackaging, but turned out this is done in PC - the client.  The crawler just provides a "source" link.)

kaste:
Since I wrote it down, it triggers me to actually provide ready-to-use packages for each package that is not a ".no-sublime-package".   Don't think that would take more than an hour or two.

kaste:
[...]

PR #96, opened 2025-09-03
Provide a web url for each release (if possible)

kaste:
Implements #88

For GitLab, it links to the generic tags page, from there you can see release notes and download the source zip.  That is probably enough.  We just don't have enough GitLab packages to justify more, e.g. to really ask the API for the release page if any.  

For GitHub, a simple regex transformation is enough.  GitHub will redirect automatically to the best page in case the tag has no release associated with it.

Bitbucket doesn't have any such site afaik.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -26,6 +26,32 @@ function minimalPackage(pkg, stats) {
     platforms: cleanupPlatforms(release.platforms),
   }))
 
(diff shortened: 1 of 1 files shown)


kaste:
We only have 1 Codeberg package and that is on a third party channel.  (Likely because we don't support Codeberg.  Which I could.  But pc.io would die.)

PR #105, opened 2025-09-15
Omit upgrade line and points if there aren't any

### LINKED ISSUE #102: Omit upgrade points for new packages
E.g.

Image

Image

There is maybe no point in showing these dots here.

kaste:
Fixes #102

That would be:

image
=>
image


braver (approved):
Better 👍🏻

braver:
cherrypicked into #107

Issue #59, opened 2025-08-03
Toggle to list layout for search results


braver:
Add the possibility to present search results in a flat list, as we do on the libraries page. Users should be able to switch between grid and list, and we should remember their preference. The static homepage with the newest and recently updated sections should probably remain in a grid regardless. The goal is mostly to make it easier to find things since you only have to read in one direction. Additionally, the list layout would create space for the package description to be displayed here as well, perhaps alongside other metadata like the last update date.

Issue #36, opened 2025-07-02
Short package description on package listing


Apis035:
Browsing on packagecontrol.io has the package short description showed in the package lists, this helps to understand what each package does at a quick glance.

braver:
This would probably go well together with changing/toggling the results to a list layout, as we’re discussing in various other issues and prs.

PR #106, opened 2025-09-15
Back'n'forth navigation for the recent section on the homepage

kaste:
On the main page, add back and forth navigation to the recent section.  This can be added to "newest" as well.

The main insight was that we actually ship -- because we have the instant search -- all data to render more recent (or newest) cards.  So we ship these 9 cards as the initial payload and then wait for the index.json (which powers the search) and then we render the actual navigation which renders the cards.  

Because I'm a programmer I use inline styles, nobody likes my cascading styles anyway.   That is also lightning fast to develop.

image
image


kaste:
Some kind of timeline indicator would ne nice.  Example:

image

kaste:
Stash some pictures 

Layout (harmonic)

image
image


Layout (ill formed metadata)

image
image

kaste:
The mobile CSS/boxes, not sure how to do that, absolutely no idea.

image

Other than: Ship it!  🌊

michaelblyons:
I think I understand, but I'm not sure. Are you saying that if I click "More," then click one of the new packages, then click "Back," you'd want to see all the packages listed that you had seen when you left? That is a great point.

I had been more considering the previous-history side: If I was on LSP's package page, then click to homepage, then click "More" 3 times, then I want to go back to LSP... I have to click "Back" 5 times to get through each extension of the Newly Updated list. That would be annoying to me.

Maybe if there is a "More" link (and it's fine if you go a different direction instead), the history can collapse to the maximum number of shown items.

braver:
What I would propose is somewhat similar to the old site. You can click "Popular" on the homepage to list all packages by popularity (https://packagecontrol.io/browse/popular), etc. Maybe less "cool", but I would much prefer the UX. It's also consistent with search and we can even easily layer filtering into that. For example, "I've got all the newest packages here, but now give me all the newest color schemes" is something that's then trivial for a user to understand how to do.

braver:
A and B are not mutually exclusive so I guess we can do both. I do now understand your ideas for B better and am more on board with it.

> The UX, (A) is a bit like an easter egg as it is not discoverable or expected that the whole page changes on that button

It doesn't need to be an easter egg. The issue with it currently is simply that I can't figure out how to make it technically. Once it's technically possible, I can add some UI to make it very obvious. That goes for "revert to normal homepage" as well, it doesn't need to be implicit. The idea of it all happening when you interact with sort is just something that came up naturally for some users, which would be the initial MVP for the feature, but I expect to be able to design features around it to make it explicit.

> The search layout can't show the date ranges right now which is an important aspect so that would need a solution

We still also want to switch search results to a list (#59, #36) instead of a grid, and I image we can then insert some strings in there to indicate months or years if you're sorted in a way that that makes sense (somewhat like [this example](https://www.muziekgebouw.nl/nl/agenda?start=2025-11-01&end=2026-01-31#anchorFilters)).

> I also totally understand the old top 25 page ("Popular"). 

"Popular" was a bad example from me, but take our headings "newest" and "recently updated", those could be clickable to list all packages sorted by that aspect.

braver:
> I'm not the manager so that is just an idea of a plan.

Well it’s not a bad plan (it’s a good one actually 😉), but I would like to have an opportunity to actually review the code and perhaps test it, before we merge. 
Also, I need a moment to read and process everything that’s written here now. 

So if you’re saying this PR is “done” technically give me a few days to give it the attention it deserves, but indeed let’s move forward as you suggest.

braver:
@kaste while the feature works well and is shippable from that aspect I would like to keep the the CSS in the CSS and there are some layout issues that I would like to resolve, before it's merged. It's harder to do it afterwards and keep it from conflicting with other potential new work. I can pick that up over the coming days if you want.

index.njk
```diff
@@ -83,4 +82,6 @@ date: Last Modified
   }) }}
 </template>
 
+{% inline_js '_includes/recent_waiter.js' %}
```
braver, line 85: Note that other inlined JS is now also in the static directory, not the _includes dir.

braver:
Actually, I would expect the same thing would work for "Newest", it would be useful to be able to browse back a bit there too I think. Any reason it's not the same for both sections?

index.njk
kaste: Then these are now exposed on the web site.  Why is that?  

Anyway, this here is inlined because it is running in a blocking manner to prevent render flicker on page load.  E.g. when you go back from a package page to the homepage.  (And you used this widget here before.)

static/module/recent-pager.js
```diff
@@ -0,0 +1,301 @@
...
+    container.className = 'recent-pagination'
+    container.style.cssText = 'display:flex; align-items:center; align-self:end; gap:0.5rem; font-size:1.3rem;'
+
+    const controls = document.createElement('div')
+    controls.className = 'button-group'
+    controls.style.cssText = 'gap:0.2ex;font-size:24px'
+
+    const first = document.createElement('button')
+    first.className = 'button'
+    first.textContent = '«'
```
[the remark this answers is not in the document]
kaste: That means we don't use fonts/font features but svg.  Is that correct?  We're here in *recent_pager*, the template in *recent_waiter* has no special chars.

These here are just the three buttons, this is done in the spirit of "progressive enhancement".

_includes/recent_waiter.js
```diff
@@ -0,0 +1,70 @@
+;(function () {
+  const RECENT_QUERY_PARAM = 'recent'
+  const RECENT_PER_PAGE = 9
+
+  const template = `<div class="card">
```
kaste, line 5: This template here is a nice template to export actually but it only holds a placeholder card.  Looks nice, has no content, and maybe could get a fixed height to match the usual height of a rendered full card better.

static/module/recent-pager.js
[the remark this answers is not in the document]
kaste: I do see the chevron left but that here is like a double-chevron.  https://primer.style/octicons/icon/move-to-start-16/?

index.njk
braver: > Then these are now exposed on the web site.

Well there is no harm in that. Nothing in the static dir is used anywhere until it’s linked, and I figured inlining them is just another way of using a script. Perhaps the inline scripts should have their own subdir, that would make things more obvious. But just letting you know where all the other ones went so this PR can align with that.

_includes/recent_waiter.js
braver: Any reason you did not re-use the template element that already exists? Or perhaps we replace that with this?

static/module/recent-pager.js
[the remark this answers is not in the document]
braver: Scherm­afbeelding 2025-09-20 om 09 29 23

There is still that misalignment though because the characters sit on the baseline of the font's character box, not in the center. These look great without that grey background rectangle behind them though, that might be a simple solution to that.

static/recent_waiter.js
```diff
@@ -0,0 +1,70 @@
+;(function () {
+  const RECENT_QUERY_PARAM = 'recent'
+  const RECENT_PER_PAGE = 9
+
+  const template = `<div class="card">
```
kaste: This is a real placeholder card.  It looks nice but doesn't have all the branches/elements we actually have.  The card-template has all the possible slots/div/spans we can eventually fill out with data.  I couldn't use card.js with it as that is not yet loaded.  There might be ways to reduce the card template by removing element branches from it, but really you could design any placeholder, just something that looks nice.  The fancy sites usually animate them to mimic a waiting state...

kaste:
Surprisingly sophisticated implementation.

braver:
> Surprisingly sophisticated implementation.

Oh yes 👌🏻 Love how you handled the history there. I had questions originally, but this already answers them. :chefs-kiss:


Sorry if this is the wrong timing to hit you with pixel-peeping but I notice these things immediately (occupational hazard I'm afraid) .. the navigation is just slightly below baseline.
Scherm­afbeelding 2025-09-22 om 20 09 12

Again, not sure if you're ready for this kind of feedback, but since we're here: the sticky thing doesn't reliably work on my machine (MacOS 15.7 Safari 18.6 or Chromium 139.0.7258.154). I did see it happen once but after several refreshes can't get it to appear again. 

All that aside though, looks amazing and putting the date ranges in there is really inspired. Nice!

braver:
> It is still PE, and as long as it loads it does not enhance the component. It is not sticky before you see the (new, added) navigation.


https://github.com/user-attachments/assets/ff5f09ab-a2eb-47bd-ae9f-fb2db28972fe

It's enhanced but no sticky. Dunno why!

kaste:
Yeah, don't take my styling as "I like it that way", it was very, very hard to get it into that so that it doesn't completely fall apart.  I think the problem started with the placement of the month-label, and the base- or visual centerline of that label and the chevrons.  Also, the double-left-chevron initially was an "x" actually. Like in: revert to the original state, close the widget.  As I thought about this as adding an "aside" or "widget" (attached) to the list, it wasn't on the header line conceptually but just near the corner of the list.    

All *this* whitespace:

image
image


E.g. right chevron, its box as highlighted, to white box of the top most card.  E.g. distance right side label to left side double-chevron, to left-chevron.  E.g. distance between the two chevrons equals or not equals distance to the cards.  🫤 😖

Issue #101, opened 2025-09-15
Net vs. Gross installs hard to read in dark mode


michaelblyons:
It's harder to see the contrast in the stat chart bars between "installs" and "installs minus removals" on dark mode than on light mode. I'm not sure if the colors are actually the same or different from light mode, but perceived contrast on dark mode is very low.

michaelblyons:
Oh, weird. You have a line between them? I was going to suggest that as a possibility. I don't see that line.

<img width="800" height="172" alt="Image" src="https://github.com/user-attachments/assets/26d44569-830b-493b-8d41-9eb12ffc76b3" />

Whoa, I can see the lines in my download and upload here, but not on the website. Same monitor for both! I don't know what's going on.

braver:
✅ via #107

PR #107, opened 2025-09-17
Chart iteration

### LINKED ISSUE #102: Omit upgrade points for new packages

braver:
Incorporates #105 and thus also fixes #102, as well as #101.

Scherm­afbeelding 2025-09-17 om 13 40 16

- increase vertical size of the chart
- move axes to the top so the grid etc don't cut into them
- improve contrast
- change the hovers (and make them work in Safari etc)
- float the chart next to the package metadata if there is space
- add a legend

The hovers now highlight the week you're on, which matches the title that's displayed. Hovering the right axis highlights the upgrades line. Hovering the legend highlights the matching chart elements.

I must say that I find the upgrade stats a bit hard to believe. A package that has been upgraded halve a year ago that sees on average 70 upgrades per week? As I understand it that means every week 70 users start ST for the first time since May, or have auto_upgrade off and manually request upgrades (who does that?)? Even at > 100k total installs that number feels high to me. 🤷🏻‍♂️ Not saying that it's wrong or broken, but it's not what I'd expect.


braver:
Good points! We do actually use net installs everywhere and simply call it “installations”. I was thinking we could perhaps do that here as well, and call what’s now labelled “removals” “installations before removals”? Because “net installs” probably doesn’t mean anything to you unless you’ve been part of these discussions, i.e. even that needs explaining. The net installs number is indeed also elsewhere but you don’t know those were “net” instead of simple all installs because it’s never explicitly specified yet.

michaelblyons:
On the subject of the X-axis direction, I want to start a list of benefits for each side.

### Descending order (current)

- The "interesting" (recent) numbers are close to the rest of the info on the page, so you don't flick your eyes as far to the right to see new data.
- PC.io does it this way

### Ascending order

- This is the normal direction of charts
- If we want a no-axis-label mini chart sometime in the future like NPM, this is the only option

### Conclusions

There's not *really* a best answer here. Either way violates the Principle of Least Surprise for some users, whether it's people used to PC.io or *\*cough\** everyone else in the world.

I had been slightly on the side of status quo, but I'm now slightly on the side of ascending order.

braver:
> There's not really a best answer here

No, there doesn't seem to be an easy decision here. Maybe move that to an issue or discussion if you want to dig in, I'd like to keep that out of scope of this iteration.

Scherm­afbeelding 2025-09-17 om 21 15 34

I did 2 things here. 

The first move is to revert back to using the gross install count as what we display primarily. It's a higher number, which is never bad :sunglasses: More importantly, we don't have removals from before our new stats, meaning the base install count for most packages that have existed before August is a gross number. In other words, dropping that number by some recent removals is slightly weird, inconsistent. Furthermore, as we've been discussing, I'm not sure I fully trust that removal number to mean what we think it does. This also fixed an untruth if you will in the UI where it previously said "installed X times": it was actually installed more than X, it was just also removed a few times ... but that's not what that says.

Then, in the chart and elsewhere things become easier to explain. We don't have to say all the time "it was installed X, but actually Y etc". No, it was simply installed X and in the chart you can see the net install being lower and you get an idea about how much, but the exact numbers aren't super relevant and we can't really talk about "total net installs" for packages from 2012 anyway.

kaste:
Love it, absolutely.  The look of the legend is great too, but the info...  

"recent average" is iirc 53 weeks.  Maybe just "Installation: 53 per week, 120.000 in total".  We can still switch to recent 4 weeks window which is maybe what we wanted.  That's how I remember it.    

Do I still have the "ceil" filter checked in?  Then: `avg_inst | ceil | grouping` to not have "74.25 installs per week".

"in total" is still duplicated, esp. when the chart jumps to the left.   

gross installs "it's the more "real" number". How I remember it: we wanted to remove the old stats and only keep the "real" numbers (ours, the more recent ones, ditching the ST glory decade).  1M installs without removals, well these just deleted ST or never installed it on their new computer.  

The average lines: I think they should be in two colors.  Usually they overlap with one portion of the bars:
image

Obviously the upper average goes through the upper bars, and the lower one goes through the darker orange one.  That's the reason I had two colors.  (Without any tuning, too low contrast... of course)

The left grey border... I don't see the grid, I don't know where that aligns to.  I actually think it floats okay if you don't draw that border. 

image

... as long as it sticks on the right edge.   What grid are you seeing here actually.

Actually I *love* the look.

kaste:
Really, I would just drop the left border and ceil the average week and ship it.

braver:
I couldn’t actually see the two different colors for the averages, the lines are too thin, but will have a look to see what can be done.

> Then: remove the original downloads counter and move the stars counter to a new place?

There’s some other potential info in that space too (“unmaintained” etc) so I couldn’t figure out what to do. I don’t really mind repeating the install number, one is just the number with an icon, the other is an explanation.

> we wanted to remove the old stats and only keep the "real" numbers

That, or put a greater emphasis on the average rate of installation because that’s more truly “current popularity”…. We should perhaps be able to sort on that rate eventually. Anyway, a topic to revisit some more!

kaste:
Naturally:  "72 per week, 7890 in the last year" (the totals of the chart we are showing) -- if we just had the numbers.  

>  the lines are too thin

They were only visible on hover (but also in my 100 pics, you 🙈) which apparently only worked on Chrome.

PR #120, opened 2025-09-22
Add magnitudes of 2.5 to the nice numbers

kaste:


eleventy.filters.mjs
```diff
@@ -115,23 +115,21 @@ function compute_step(arr, target) {
...
-  if (lead_number <= 1) {
-    return 1 * mag
-  } else if (lead_number <= 2) {
-    return 2 * mag
-  } else if (lead_number <= 5) {
-    return 5 * mag
-  } else {
-    return 10 * mag
+  let normalized = approximation / mag
+  const niceSteps = [1, 2, 2.5, 5, 10]
```
kaste, line 126: Much easier to read than before.

braver:
Much better!

This is just one case where it goes to 50 and the chart's kind of stuck halfway. But as I think you've explained, you can't really improve that without doing what D3 does and let go of the "predictable" axis ticks. I don't mind this so much, in most cases it's quite a bit improved by this PR 👍🏻 

Screenshot 2025-09-22 at 14 40 27

braver (approved):

PR #121, opened 2025-09-22
Tune Minisearch with a custom tokenizer

### LINKED ISSUE #46: Search fuzzy find, camelcase, and spaces
There are some issues with the search not finding things I'd hope it to:

- Camelcase substrings are not returned. `cmd` should find `SublimeCMD`, right?
- Searching for split words doesn't find those words together. `you complete me` can't find `YouCompleteMe`.

Discovered during the package sorting for #40.

kaste:
Add a harmless tokenizer without ngrams just to *also* split camelCases. Turn off fuzzy because it doesn't help us, e.g. it matches "rsp"ec for "lsp".

Fixes #46

### DIFF
--- a/static/module/minisearch.test.js
+++ b/static/module/minisearch.test.js
@@ -33,6 +33,34 @@ describe('GitSavvy search', () => {
       platforms: ['linux'],
       labels: ['color'],
     },
+    {
+      name: 'LocalisationHelper',
+      description: 'Handles localisation workflows',
+      author: 'Locale Co',
+      platforms: ['linux'],
+      labels: ['localisation'],
+    },
+    {
+      name: 'LocalizationHelper',
+      description: 'Handles localization workflows',
+      author: 'Locale Co',
+      platforms: ['linux'],
+      labels: ['localization'],
+    },
+    {
+      name: 'InternationalisationSuite',
+      description: 'Supports internationalisation efforts',
+      author: 'Global Co',
+      platforms: ['linux'],
+      labels: ['internationalisation'],
+    },
+    {
+      name: 'InternationalizationSuite',
+      description: 'Supports internationalization efforts',
+      author: 'Global Co',
+      platforms: ['linux'],
+      labels: ['internationalization'],
+    },
   ]
 
   let minisrch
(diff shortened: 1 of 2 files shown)


braver:
The fuzzy intends to help against minor typos, like also I guess for instance cases where US English uses a z and GB English an s,  colour vs color etc.. In other words to be accepting of small user errors. Those near misses should be scored lower than correct hits, so with the magic sorting the cases where you got it exactly right should be higher up in the results. That's to say that I would hope that a small amount of fuzzy adds to the user-friendliness of search and should not cause problems. But perhaps it doesn't work as well as I hoped? 

Splitting camelCase is obviously a win of course, nice 👍🏻

kaste:
fuzzy is a bit misleading when we come from Sublime Text.  It doesn't even do substring search nor camelCase.  I think it is not user-friendly here as we often see results and ... "Huh?  What does match here?" like rafal (IIRC) pointed out on discord.  

If you search "gti" and it doesn't find "git", you probably find out;  but right now you search "js" and it doesn't find "AlpineJS" IIRC.  

For our lucky case, only the searcher might have typos.  Unlikely, the data has typos in it (which is not really always the case when you implement searching).  

We should add color/colour as that is important in our domain.  Do you have other weasel words?  Try color/colour in the current deployed version; and compare to pc.io.  (pc.io does this correct, that is not fuzzy typing)

kaste:
Ok.  This is now test-driven.  The tests are actually easy so we can add new behavior at any time.

static/module/minisearch.test.js
```diff
@@ -33,6 +33,34 @@ describe('GitSavvy search', () => {
...
+      platforms: ['linux'],
+      labels: ['internationalisation'],
+    },
+    {
+      name: 'InternationalizationSuite',
+      description: 'Supports internationalization efforts',
+      author: 'Global Co',
+      platforms: ['linux'],
+      labels: ['internationalization'],
+    },
```
kaste, line 63: 😂

braver:
@kaste I did feel that I fully understand the setting, although I did initially just take the value from their docs. 0.2 means you get to make 20% typos: in a 5 character search term, 1 typo is allowed. That would mean that "convertor" would still find "converter", while the ranking should still score any hits on "convertor" higher. It's not fuzzy like Sublime's search is fuzzy (where "cvrtr" would find both "convertor" and "converter"). And that makes it probably not fuzzy enough to be of any use. I.e. it does _something_ that I think might be useful, but it has the downside of serving a lot of false positives and unexpected results. So, I'm not sad to see it go. I think if we really care about fuzzy searching we'd be better off finding a different library rather than messing with this setting.

kaste:
The downranking is maybe even the problem here and not the solution.  If I write convrter, I don't want any "downranking".  Either it understands that as a typo or not.

PR #122, opened 2025-09-22
Switch to search mode when interaction with the search options

kaste:


static/module/search.js
```diff
@@ -73,4 +73,14 @@ export class Search {
       score,
     }))
   }
+
+  // return all indexed packages using minisearch's wildcard query
+  all() {
+    this.stringSearch = false
+    const wildcard = this.minisearch.constructor?.wildcard
+    if (!wildcard) {
+      throw new Error('minisearch wildcard symbol is unavailable')
```
kaste, line 82: LOL, of great help of course.  For the end user!  🏩

static/module/list.js
```diff
@@ -151,8 +158,6 @@ export class List {
     // render results with pagination
     this.renderPage(sortedResults, page)
 
-    this.section.dispatchEvent(
-      new Event('search-is-ready', { bubbles: true }),
-    )
+    this.section.dispatchEvent(new Event('search-is-ready', { bubbles: true }))
```
kaste, line 161: Looks like I fired Prettifier up.  Mhm.

static/module/search.js
braver: where is a `sublime.status_message()` if you need one ;)

static/module/list.js
```diff
@@ -120,28 +124,31 @@ export class List {
     // Only push state if URL is actually changing
     if (window.location.search !== queryString_) {
       const target = '/' + queryString_
-      const title = value.length > 0 ? `Search — ${value}` : this.initialTitle
+      const title = hasQuery ? `Search — ${query}` : this.initialTitle
```
kaste, line 127: It already pushes to history but here it doesn't change the title.  Obviously, I had no query to show.  but not changing the title but pushing is not an option ...  and not pushing is not correct either ... so we need a title.

braver:
Very nice actually. Resetting by putting it back to "magic" seems reasonably logical to me. And one can also click the brand in the top left. 

One interesting thing I noticed: sort by author (http://localhost:8080/?sort=author):

Scherm­afbeelding 2025-09-22 om 20 19 30

... what to do with all those packages that don't have an author? 🤔

kaste:
- Special case missing authors when listing by author
- Do page titles

kaste:
This can likely go.  A lot of refactoring went into index.js

static/module/list.js
```diff
@@ -116,32 +132,47 @@ export class List {
...
-    // Only push state if URL is actually changing
-    if (window.location.search !== queryString_) {
-      const target = '/' + queryString_
-      const title = value.length > 0 ? `Search — ${value}` : this.initialTitle
+    const target = queryString_ ? '/' + queryString_ : this.initialPath
+    const currentPath = `${window.location.pathname}${window.location.search}`
+    const sortTitle = this.sortTitleMap[sortBy] ?? sortBy
+    const title
+      = usingWildcard
+        ? `List by ${sortTitle}`
```
braver, line 140: Teensy tiny minor point: currently this becomes a kind of title case ("List by Newest"), where I think it might look more natural if the sortTitle's first character is lowercase ("List by newest").

Other than that: quite nice 👌🏻

kaste: It is called title case for a reason.  `const title = ...`

braver: Fair enough 😄 I don't have a strong opinion, if you prefer it as is let's go with that.

braver (approved):

PR #125, opened 2025-09-23
Avoid flicker when reloading a search

kaste:
On the homepage we now have our brand in an `<h1>`.  Do not hide this when reloading a search page.


braver (approved):
Good catch!

Issue #124, opened 2025-09-23
Top section in details page not necessarily tall enough for the stats


kaste:
<img width="1338" height="629" alt="Image" src="https://github.com/user-attachments/assets/86519fd5-b3b5-4daf-a02e-aa05c2ef86a1" />

braver:
Fixed in #126

Issue #123, opened 2025-09-22
Card box error for a RIP package


kaste:
<img width="406" height="81" alt="Image" src="https://github.com/user-attachments/assets/2e8aceec-0c95-4d2a-bf0c-29cd545aa823" />

braver:
Fixed in #126

Issue #115, opened 2025-09-21
Average lines color is not set against the bars colors


kaste:
<img width="97" height="189" alt="Image" src="https://github.com/user-attachments/assets/6a0bd1d6-2d7e-4591-939c-ecd7100e4dd0" />

<img width="114" height="184" alt="Image" src="https://github.com/user-attachments/assets/9b740592-d063-4e1e-8c05-4b3c1415eb13" />

I thought actually part of #107.  The lines have two colors now, and the upper contrast's well against the bars, esp. the upper where it mostly is or should be.
But the lower (net) average has basically no contrast to the bars, neither the upper bar nor the lower bar.

braver:
Fixed in #126

PR #112, opened 2025-09-19
Search results in list not grid

braver:
- fix #59: display search results in a list
- fix #36: display package description in search results

I decided against making a toggle (for now) because it adds complexity and we've gotten ample feedback that for finding, discovering and browsing packages a list is preferable. On the homepage the 3x3 grids are nice and compact, but when you're trying to find things going through a list in one direction is simply easier.

We already have a list layout (for libraries) so it only took some minor tweaks.

I don't love how the entries look spread out over 1200 pixels wide, so I limited the width and entered the list. Takes some getting used to, might feel different about it after I sleep on it.

Considered displaying the last-modified date in there too, since we have it and in chronological lists might make sense. Maybe that needs to switch to creation date if sorted by that aspect? I wasn't sure about this so reverted for now. This is a screenshot of full width with that date:
Screenshot 2025-09-19 at 15 28 56

This is current state of the PR:

https://github.com/user-attachments/assets/17fe29fa-27fc-4e8e-884e-8641b14b8a1d

Easy peasy :)

PR #126, opened 2025-09-24
Assorted tweaks

braver:
Adds polish to features we've shipped the past few weeks, and switches search results to list layout (#59, #36):

image

- #123 
- #115 
- #124 
- #59 
- #36

- minor layout polish on the new pager
- fix my previous work (#112) so that the pager doesn't accidentally add descriptions on the cards
- fix that same work so it doesn't break the libraries pages
- fix contrast on the chart (average and upgrade lines) and thereby improve color contrast overall in dark mode
- improve line length for readability everywhere and more consistently

This is mostly just very quick work yesterday and today, but I think I made no major mistakes 🤞🏻

### DIFF
--- a/libraries.njk
+++ b/libraries.njk
@@ -21,7 +21,7 @@ date: Last Modified
     {%- endfor %}</p>
 
     <p class="description">
-      {{ pkg.description }}
+      {{- pkg.description -}}
     </p>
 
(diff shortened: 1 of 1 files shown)


braver:
Seriously, it's _so good_ 😄

PR #128, opened 2025-09-24
Fix: Guard against homepage being null

kaste:
After an update to thecrawl, we no longer store fallback homepage values in the workspace.json.  (thecrawl does that in a later step for the channel.json only.)  In workspace.json it was just a duplication of the details value.


michaelblyons:
Is this the thing where "Homepage" link on package details just links to the page you are already on if it's not provided in the PCC json? That also seems to be true for the href of ReadMe.

If not, I can open a new issue.

PR #137, opened 2025-10-01
Labels page

### LINKED ISSUE #132: Consider creating a "Labels" page
In other words, the equivalent of https://packagecontrol.io/browse/labels. It's not super useful for most users I imagine, but I use it from time to time to check when someone submits a new package with an uncommon label: it's a quick check to see if it's used more often. 

Although I could easily accomplish that with a quick search, e.g. https://packages.sublimetext.io/?q=label%3A%22powershell%22. 

The question then is I guess if a Labels page is useful or interesting for discovery, or insight into the most popular labels. The answer could easily be "no" and then we close this issue 😄 But I wanted to bring it up as it would make the new site "feature complete".

braver:
Fix #132

That felt easy. Minimal new UI. I just link to this page in the footer... no need to call special attention to it IMO, and of course there isn't quite enough space below the search field next to the labels buttons.

Scherm­afbeelding 2025-10-01 om 16 37 46

It's immediately useful for catching odd mis-labelings:

Scherm­afbeelding 2025-10-01 om 16 07 39

... and incorrect case (convention is all lowercase):

Scherm­afbeelding 2025-10-01 om 16 07 44

There are also various other cases where the number of times the exact label occurs doesn't match our search results for that label (because the search is more lenient). Try [search](http://localhost:8080/?q=label%3A%22search%22) for instance: there are 34 occurrences but our search finds 38 packages... not sure what to do about that.


michaelblyons:
>>  search must push to history/URL to be fully compliant
>
>Good point, that's missing from the libraries page too.

Since you can't link directly to a page for a specific library, this will be helpful for sharing links with other people.

kaste:
If we only do one thing, we should do By Name IMO.  That doesn't mean the other thing has no use case.  

If you imagine on the home page the "..." and here the By Occurrence, then naturally the 5 labels on the homepage would represent the top 5.  I don't think they are at the moment and they shouldn't, again IMO.  That is an indicator that labels maybe shouldn't be ranked.

braver:
Got it to fit!

Scherm­afbeelding 2025-10-02 om 20 59 10

michaelblyons:
If it's pixel perfect, are there going to be problems with rendering on different platforms?

PR #139, opened 2025-10-02
Teach the search buttons to act as toggles

kaste:
I.e. when you click on "linting", we set "label:linting".  If you click
"linting" again we unset the filter.  *Changing* the label works as
before.

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -27,6 +27,74 @@ const form = document.forms.search
 const input = form.elements['q']
 const sortSelect = form.elements['sort']
 
+/**
+ * @typedef {Object} FilterButtonDescriptor
+ * @property {HTMLAnchorElement} element The original filter button element.
+ * @property {string} type The token prefix (e.g. label, platform, author).
+ * @property {string} token Full token string extracted from the button href.
+ */
+
+/**
+ * All filter buttons that can toggle search tokens, converted to metadata objects.
+ * @type {FilterButtonDescriptor[]}
+ */
+const filterButtons = Array
+  .from(form.querySelectorAll('.button-group.labels a.button[href*="?q="]'))
+  .map(parseFilterButton)
+  .filter(Boolean)
+
+/**
+ * Extract token metadata from a filter button anchor.
+ * @param {HTMLAnchorElement} element
+ * @returns {FilterButtonDescriptor | null}
+ */
+function parseFilterButton(element) {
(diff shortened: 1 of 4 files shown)


kaste:
A click.  

I just made the hover thing, and then one step  further, why not remove them on click.  I actually wanted to make "x"s to click them away; but for the first draft was happy to get that hover plus click action somehow.

kaste:
That fancy UI is not part of this PR.  Also because of that: there is no click-through.  Adding a "x" should have solved that but I was happy to get it to that state actually.  

However, just the hover effect, hover over a button to highlight the term in the input box, is also shown in the video.  I slightly found that to be more elegant than to outline all matching buttons (to indicate:  these are toggled "on") which is a bit ridiculous if you search a label and all cards then have an outlined label button on them.  To resolve that, I followed the idea not to highlight each matching button, but to highlight matching query terms.  But when you have that you want to interact with them directly, hence the remove on click idea...

braver:
Maybe to clarify (with more words) my UX design perspective on this.

The labels in the cards are not the current search query, they have no relation to the search query. They are entries into new search queries using those keywords. In other words, if I search for color schemes and I see one by "braver", clicking braver should start a new query for packages by that author: it never affects the current query and it shouldn't. If there is an interesting label on any of the results that will take me down into another rabbit hole with results for that label. The cards themselves provide navigation into other avenues of search.

The search box holds the current query. We can add UI in there to make it easier to edit the search query. That makes a lot of sense because that _is_ the search query you're interacting with. It even makes sense for the buttons directly underneath the search box to dynamically update and refine the query: they're visually and conceptually all about editing your query. The labels on the cards are not your current query and are not about modifying it.

braver:
> And I didn't know if that is actually intended or how to make that more obvious that these are two different things.

It was intentional to make those buttons in the form different from the ones on the cards: they refine the queries. Although, having gone and thought a bit more about the topic via this PR.. I now think that's weird too. Or at least, those buttons don't clearly signal that they do that. The idea originally came from the 3 platform buttons: they only make sense as a refinement ("packages for query that work on Windows"), not as a query all by themselves ("all package that work on Windows"). So then I extended that to the other 6 label buttons. And it does make sense that if you see results where a lot of package have a certain label, that you want to refine by that ("but give me all color schemes for query")... but even there the UI is already lacking.

PR #142, opened 2025-10-02
Avoid form flicker in slow loading environments

kaste:
* Actually hide the form until "DOMContentLoaded" otherwise the placeholder gets rendered resulting in a fast flicker: "Search 1926 packages" -> "label:amxmodx"
* Add support for "sort" param

### DIFF
--- a/static/handle_search_url.js
+++ b/static/handle_search_url.js
@@ -7,8 +7,8 @@
 (function () {
   const params = new URLSearchParams(location.search)
   const q = params.get('q')
-  const isListing = !q && params.has('sort')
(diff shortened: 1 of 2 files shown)


kaste:
I know nobody cares about such stuff, but *I* actually noticed the flicker when testing the labels feature.

braver (approved):
> I know nobody cares about such stuff

I do care! 🤗 I just don't notice it immediately or am distracted by something else.. or don't always see a solution. 

This does look more stable if I refresh a lot quickly etc. Looks good, I'm glad you can pick these up to get the details right 👍🏻

PR #143, opened 2025-10-02
Fix card outline bleeding into the sticky header

### LINKED ISSUE #141: Card outline is clipped after pager header got sticky
The on-hover outline of the cards get clipped on the top if you make the pager-header thing stick and then scroll back up so it un-sticks.

Image

It looks to me like it retains the `.is-sticky` class even when it's not longer "stuck". Removing that class is one way to solve it. Some juggling with margins could be another solution.

kaste:
Fixes #141

### DIFF
--- a/static/module/recent-pager.js
+++ b/static/module/recent-pager.js
@@ -10,6 +10,10 @@ const RECENT_PAGER_STYLES = `
     justify-content: space-between;
     gap: 1rem;
 
+    > h2 {
+      padding-top: 0;
+    }
+
     &.is-sticky {
       position: sticky;
       top: 0;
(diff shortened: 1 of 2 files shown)


kaste:
Pick a commit  :grin:

kaste:
I think we don't want the third commit as it defeats the purpose of what an outline is.  Basically it is a border then. That still looks good but it just not what we had. The growing effect. The first commit is the hack, and the middle commit is IMO the typical solution. However, the middle commit shifts the cards down for two pixels.

braver:
I agree the middle commit is the right approach. The 2px isn't a problem, but you can remove 2px of margin on the H2 in the sticky pager-header to compensate. That looks just perfect to me 👌🏻 

```diff
diff --git a/static/styles.css b/static/styles.css
index ae9fd9b9..4dabe3e0 100644
--- a/static/styles.css
+++ b/static/styles.css
@@ -160,6 +160,7 @@ section {
     > h2:first-child,
     .pager-header > h2 {
       padding-top: 0;
+      margin-bottom: 8px;
     }
   }
 }
 ```

kaste:
`margin-bottom: 8px;` moves the whole right side, label plus nav.  The baseline?

We should also only do anything if ".is-sticky" is set, likely in "recent_pager.js"

PR #147, opened 2025-10-05
Some UI around keyboard shortcuts

braver:
Not fully implemented, not polished, and for the hint around the search binding you can pick from 2 commits. But for free, no strings attached 😉 

Like so...
Scherm­afbeelding 2025-10-05 om 15 37 25
... but 2 placeholders is weird so this could simply also work
Scherm­afbeelding 2025-10-05 om 15 36 53

And toggles
Scherm­afbeelding 2025-10-05 om 15 48 12


kaste:
Generally, I would have liked a keyboard icon e.g.

image
image

for which we indicate on/off with something like inverting the icon.  A heavy black is on, and a lighter white is off.  I know you reply with, he, how should the user know which on is which, but to be fair I also don't know it in above picture.  Are the keys on or off if the toggle is left.  (I pretty much hate these left-right toggles because I almost never know if the toggle is on or off.)  The light/dark mode toggle is perfect.  So for the keyboard shortcut we can do a toast saying on or off.

PR #150, opened 2025-10-07
Hotfix css selector to match the labels page specifically

kaste:


kaste:
Must have lost it during the refactorings, but the CSS selector was not specific enough and also matched on the search results.

PR #151, opened 2025-10-07
Add ".readme kbd" style

kaste:
The exact colors aren't even that important here.  But it turns

image

into

image

and that is clearly better.

### DIFF
--- a/static/style/readme.css
+++ b/static/style/readme.css
@@ -27,11 +27,11 @@
 
   kbd {
     display: inline-block;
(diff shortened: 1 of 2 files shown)


braver:
Nice one. Pushed some small tweaks so it also work in the dark.

Scherm­afbeelding 2025-10-08 om 10 08 00

Scherm­afbeelding 2025-10-08 om 10 08 40

🚢 !

kaste:
🌊

PR #158, opened 2025-10-15
skip to search doesn't work reliably on detail

braver:
Because it's collapsed into an icon, we can't jump focus in to the search box. This is a cheap way out. Alternatively we skip to the search button. But it's only 2 tabs away anyway, and search isn't the focus feature of the detail page.

### DIFF
--- a/_includes/header.njk
+++ b/_includes/header.njk
@@ -1,9 +1,11 @@
 {% import "util/macros.njk" as util %}
 
 <header>
-  <a href="#search-field" class="skip-link">Skip to search</a>
-  <a href="#main-content" class="skip-link">Skip to main content</a>
-  <a href="#footer-links" class="skip-link">Skip to footer links</a>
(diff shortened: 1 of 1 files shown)


kaste:
Ah.  I thought this minimal, shrink'ed design is for tablets/mobiles, and then you don't have a tab key anyway.  Do we maybe switch to that layout too soon?  I should just take a look and open, de-shrink the icon to the expanded form then.  

Now that I do this, there is an apparent visual bug here: 

image

Actually, I noticed that before but on my phone I just was lucky and the icon was placed in a pleasant spot.

braver:
If you link to an element by id that way, focus jumps there. If you link to any form field that way, you enter that field... To me it looks like that's "normal" (although I have never seen it used in the wild). So my guess would be "good enough".

kaste:
But we literally don't follow that link, there is no reload, and the URL stays the same.  🤔

PR #159, opened 2025-10-15
make active buttons appear "pressed in"

braver:
A small refinement as I proposed in #139

This uses the metaphor of the buttons being stuck, meaning you can press again to unstick.  It also leaves outlines to just mean "focused". Overall it's just a little more subtle, dare I say sophisticated 😉 

Scherm­afbeelding 2025-10-15 om 12 36 07

### DIFF
--- a/static/style/buttons.css
+++ b/static/style/buttons.css
@@ -43,9 +43,17 @@
 
   &.is-active {
     background: var(--background-4);
-    box-shadow: inset 2px 2px 2px var(--background-5);
-    &.is-active:hover {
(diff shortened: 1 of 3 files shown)


kaste:
That's brutally subtle.

image
image

 I would have thought that you also need to change the background, maybe similar to the second picture.  On the other side, better than what we have...

braver:
> brutally subtle

LOL. I don't disagree, I'll add a little more oomph.

braver:
> I would have thought that you also need to change the background

Yes, you're right. I made them darker when pressed-in now.

PR #163, opened 2025-10-17
Switch meaning of "newest" to "first_seen"

kaste:
After ingesting all the old "first_seen" data into "workspace.json", switch from "created_at" to "first_seen".

This makes the newest section on the homeland and the list by newest page match the order found on pc.io.

Also, typical:  I create a package actually a year before I eventually decide to publish it on packagecontrol.  Such a package would never appear on the homepage.

"first_seen" has roughly the meaning "registered_at" but we don't add that to the "registry.json" during registration, although it is hard to recreate on data loss.  🤷 On the package details page we still show "Created at ...".

### DIFF
--- a/eleventy.config.mjs
+++ b/eleventy.config.mjs
@@ -74,6 +74,7 @@ function basePackage(pkg, stat) {
     author: util.cleanAuthors(pkg.author) ?? [],
     stars: pkg.stars ?? 0,
     installed: stat?.installs?.totals ?? 0,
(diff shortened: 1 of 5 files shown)


braver (approved):
Very nice. I regularly see the newest - by submission/approval date - packages not at the top of "newest" and always have to remind myself that newest doesn't mean what I think it means. We used to not have first_seen though right? Or was that just for pre-thecrawl packages?

PR #164, opened 2025-10-17
Fix hover over "active" (pressed) grey button

kaste:
Specificity of `&.is-active:hover` beats `&.is-active &:hover` and that is probably not what we wanted.


braver:
Indeed!

PR #168, opened 2025-10-19
Chart iteration

### LINKED ISSUE #113: Version releases on stat chart
@kaste [had these very cool samples](https://github.com/packagecontrol/thecrawl/issues/95#issuecomment-3268295100) of new versions of a package listed on the stats chart:

image
image
image

I think that's great context, especially plotted against the Upgrades data.

I'm curious about how the following things affect display:
- Lots of releases close together
  - could be "fast follow" releases
  - could be devs who maintain separate branches for multiple ST ranges and backport updates
- Different ST range releases in general

For "too close," it's mostly a question of displaying the label without overlap. Things I've thought of:
- The site could just take the biggest semver if releases are in the same week
- Having an offset for labels that cycles between some number of values

For different ST version ranges, here are some possibilities:
- Different color dots for each range, and list them in the legend
- Just let people figure it out, since there are probably not that many packages that have a release for multiple ST ranges in the last year
- A shorter chart below that shares the X-axis with the current one and offsets each ST range's release series. (But this seems way overkill for most packages...)

I am interested in what it would look like for the display of each version (assuming not having the second whole chart) to have a dot on the X-axis and a vertical line all the way up to the top. I don't have a problem with any of the current images, but I think I often see "context event" type data as vertical bars.

kaste:
- Draw dashed line from the running week to the week 1
- Bake package updates into the chart

Fixes #113 

image
image

The third commit makes a rather arbitrary change to the circle thickness.  The white circles made it hard to read, although I prefer if they are drawn against the grey background; this one is maybe too bold.  Not my key competence, really.

### DIFF
--- a/static/style/chart.css
+++ b/static/style/chart.css
@@ -1,4 +1,15 @@
 .chart-container {
+  /* Chart color tokens (light theme defaults from latest commit) */
+  --install-bar-color: var(--orange-1);
+  --install-bar-hover-color: var(--orange-2);
+  --upgrade-line-color: hsl(212deg, 88%, 62%);
+  --upgrade-point-stroke-color: hsl(0deg 0% 82.07%);
+  --upgrade-point-fill-color: var(--blue-3);
+  --release-point-fill-color: var(--red-1);
+  --release-point-stroke-color: hsl(7deg 91.82% 41.44%);
+  --avg-line-color: var(--orange-3);
+  --install-legend-color: var(--orange-1);
+
   --spacing: calc(2 * var(--side-padding));
   max-width: calc(var(--max-content) * 2/3 - var(--spacing));
   margin: 10px 0 18px 0;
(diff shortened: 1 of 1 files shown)


kaste:
The bold circles could be the on-hover style, lol.

braver:
Not to completely take this PR off course, but ... It might just be my eyes or my devices, but where blue meets orange doesn't look right to me.

Scherm­afbeelding 2025-10-22 om 09 59 23

I think it's because the values are actually almost identical.

Scherm­afbeelding 2025-10-22 om 10 04 46

Which is by design of course, but that design didn't account for them being layered on top of each other. 

With certain kinds of color blindness that also becomes a bit of a problem.

Scherm­afbeelding 2025-10-22 om 10 05 58

... all that to say that picking a completely different shade of blue, or just not blue but something else, is also ok. Probably out of scope for this PR. We can bike-shed color choices for months 😄 

---

Did I mention that the arrow pointing to the dot is really awesome?

Scherm­afbeelding 2025-10-22 om 10 07 57

kaste:
LOL, yeah, had luck with that arrow.  I really like how direct and simple SVG is. Just minutes from idea to drawing actual pixels.  (Unlike CSS).

Guys I revert the css change and merge.   

Please keep an eye open:  if the upgrade point is exactly at the top, I don't draw a callout.  That was intentional first because I didn't want to reach outside the chart.  But now I cannot find that edge case package anymore (or these, who knows).  Original idea was to flip to South but since then the code changed and we actually draw callouts slightly above the chart area.  Likely I just need to adjust the minimum length or gaps for the edge.  I just lost my test case.

braver:
Wooooo 🤩

PR #174, opened 2025-10-22
Correctly implement the a11y features for the search icon link

### LINKED ISSUE #161: Skip-to-search doesn't work reliably on detail
Originally in #158, but that's unlikely to be the right solution after all. Reporting here as a bug report for better tracking.

Because it's collapsed into an icon, we can't jump focus in to the search box. The skip link is visible but clicking it or pressing return does nothing. Additionally, the s and / keybindings also don't do anything in this situation.

https://private-user-images.githubusercontent.com/2543659/501500799-2aa1b883-5bf3-44ae-8ed4-0264bc69422f.mov

kaste:
Fixes #161

### DIFF
--- a/static/compact_search.js
+++ b/static/compact_search.js
@@ -70,3 +70,10 @@ if (toggle) {
     input.focus()
   }
 }
+
+// Allow other scripts to request expansion without focusing the input.
+window.addEventListener('search:expand', () => {
+  stopAnimations()
+  form.setAttribute('data-expanded', 'true')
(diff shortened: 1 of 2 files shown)


static/style/search.css
```diff
@@ -167,7 +167,7 @@ form.compact {
     }
   }
 
-  &.has-attention {
+  &[data-expanded="true"] {
```
kaste: This is like a simple rename "data-expanded" just has the standard collapsed/expanded notion.  I literally already forgot what "has-attention" actually means, and I think it is exactly that.

PR #175, opened 2025-10-22
Add "http-server" and usage recipes

kaste:
Using `http-server` has the advantage of behaving more like production. In particular, it enables bfcache which is important for our page because it restores the scroll position on back/forth navigation which we rely on.

(The pages are also served *while* building FWIW.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,38 @@ A Sublime Text packages database
 - [the crawl][db] database on [the `main` branch][thecrawl].
 
 
-## Running locally
+## Run locally
 
-```sh
-make build
-make serve
+You have multiple options. After
+
+```bash
+npm install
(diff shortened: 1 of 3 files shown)


kaste:
Moderately nerdy.  I did not add this to Makefile as I can't run Makefiles.  (Except when I'm on Linux.)  The `npm run x` is also just simpler because npm resolves the executable.  Not a tooling expert here, but having it in package.json is maybe still "the right thing" because of that.  "npx" ist just slow.  Every day (or half a day) it just thinks for tens of seconds.

braver:
Cool 👍🏻 Feel free to ditch the Makefile if you're not using it, two sets of tools is perhaps too much anyway.

PR #176, opened 2025-10-22
Implement "browse/new/rss"

### LINKED ISSUE #171: Provide RSS feed with new packages
This is currently used in the ⁠[package-announcements channel](https://discord.com/channels/280102180189634562/280157209017450507).

It should provide title, author, description and a link. 

This is the current one from the old site: https://packagecontrol.io/browse/new/rss

It should also be linked in the head like [so](https://learn-eleventy.pages.dev/lesson/17/#adding-the-rss-meta-tag) for instance:

```html
<link rel="alternate" type="application/rss+xml" href="{{ site.url }}/feed.xml" />
```

kaste:
Fixes #171

### DIFF
--- a/_headers
+++ b/_headers
@@ -1,0 +1,2 @@
+/browse/new/rss
+  Content-Type: application/rss+xml; charset=utf-8
--- a/_includes/head.njk
+++ b/_includes/head.njk
@@ -17,11 +17,12 @@
 
   <link rel="stylesheet" href="{{ '/static/styles.css' | bust}}">
   <link rel="icon" href="/assets/favicon.ico">
+  <link rel="alternate" type="application/rss+xml" href="{{ site.origin }}/browse/new/rss" title="Newest Sublime Text Packages">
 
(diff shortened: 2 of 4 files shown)


kaste:
That's seems very straight and was trivial to do.  Let's see if it works.

kaste:
Added a so called Transformation Rule to CF that rewrites the response header to "Content-Type: application/rss+xml; charset=utf-8".  That looks correct.  I actually don't have a RSS Feed reader at hand to check.  A next patch would be to only add that to `<head>` on the homepage, or on the "List by Newest" page.

kaste:
I asked Terminal to give the PRO to a public account (like "Sublime-Community-OSS" whatever).  Currently, the pro features are sitting on his personal account for which he won't or can't give access to anyone.

kaste:
We can still build on Github.  Just the final step would be to deploy/push it onto CF.  For that we need an API token for which we need Account rights.  Without the OSS sponsor, we would need to spend 5$/week for what we currently have if I'm not mistaken.  Certainly not arms and legs.

PR #180, opened 2025-10-24
Add `human-date:focus-visible` style

kaste:


kaste:
IMO this should be refactored so that it is in one place, e.g.

```
a.focus-visible, button:f-v, [role=button]:f-v {...}
```

PR #186, opened 2025-10-25
Implement search state in the URL for the libraries and labels

### LINKED ISSUE #146: Add browser history for labels and libraries searches
Labels and libraries pages have very simple search where the elements already rendered in the page are toggled based on the search terms. Browser history is not affected at all and these pages don't accept query parameters in the URL. 
Ideally `/labels/?q=lsp` and `/libraries/?q=lsp` would actually search for "lsp" on those pages. 

Note that some anti-blinking measures should probably be taken just like on the main search. 

And if whoever takes this on is feeling super inspired, it might be nice to look into adding a sort option for the labels page, to sort by count.

kaste:
Fixes #146

### DIFF
--- a/static/handle_search_url.js
+++ b/static/handle_search_url.js
@@ -13,8 +13,27 @@
   };
 
   const root = document.documentElement
+  const isHome = root.classList.contains('home')
+  const titlePrefix = (() => {
+    if (isHome) {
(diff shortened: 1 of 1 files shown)


braver:
Pfew! 😅

PR #194, opened 2025-10-28
Fix `release.web` links for GitHub

kaste:
We were too restrictive.  This is valid too:

`https://codeload.github.com/michaelblyons/SublimeSyntax-HTML-CSharp/zip/version/st4157/5.2.0` ->
`https://github.com/michaelblyons/SublimeSyntax-HTML-CSharp/releases/tag/version%2Fst4157%2F5.2.0`


kaste:
@michaelblyons ☝️ That is your package.  You reported it at some point, but there is no issue for it open.  Is that correct?

FichteFoll:
Wait, SM supports tab grouping when you use slashes? That is an interesting piece of information that I might consider in the future.

michaelblyons:
> Wait, SM supports tab grouping when you use slashes?

You get a collapsible tree view like you do with branches.

PR #196, opened 2025-10-28
Fix: Draw callout for release dots at the upper chart edge

kaste:
I reported that in the original PR, and LSP exposes the bug.

image


kaste:
image

That's the minimal patch to support it.  I tried small lines facing North but that didn't look good.  In the end, we don't draw a line here and just have 6px gap from point to text.  (Default gap is 4px.)

We can't go E or W as easy as we don't compute the available space for these directions.  

The original idea was to go S and then really to draw more like a box with a background and text in it.

PR #199, opened 2025-10-28
Implement time indicators interwoven with the cards

### LINKED ISSUE #135: Indicate time period in chronologically sorted results
It would be helpful to have some sense of time in lists (search results) sorted by date. E.g. when browsing back through recently updated, "are we in 2021 yet"? The pagers on the homepage does that very elegantly, we could replicate that in search results. That wouldn't add markers inside the list ("2024 starts here") but rather indicate the spread of the current page.

Perhaps, like on the homepage above the list:

Image

Or near the pagination:

Image

These are just manipulated screenshots by the way.

kaste:
Fixes #135

We split the one big ul and interweave it with `p.timeline-break`
elements.

In general, we iterate over the contents of the whole page

```
    pageItems = this.pagination.calculate()
    buildTimeline(pageItems)
```

and find a bucket size, either "monthly", "quarterly", or finally
"yearly".  We aim to split not into more than 3 groups, unless we're
in "yearly" mode.

### DIFF
--- a/static/styles.css
+++ b/static/styles.css
@@ -244,21 +244,15 @@ section {
 }
 
 .search-results-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 1rem;
   margin-bottom: 4px;
 
   > h1,
   > h2 {
     margin: 0;
-  }
-}
-.pager-header {
-  display: flex;
-  align-items: center;
-  justify-content: space-between;
-  gap: 1rem;
-
-  > h1,
-  > h2 {
     padding-top: 0;
   }
 
--- a/_includes/search_section.njk
+++ b/_includes/search_section.njk
@@ -1,6 +1,6 @@
 <section data-list-target="section" style="display:none;">
 
-  <div class="pager-header search-results-header">
+  <div class="search-results-header">
     {% if isPackagePage %}
       <h1 data-list-target="heading">Results</h1>
     {% else %}
--- a/static/module/recent-pager.js
+++ b/static/module/recent-pager.js
@@ -5,6 +5,15 @@ const pagerRegistry = []
 
 const RECENT_PAGER_STYLES = `
   .pager-header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 1rem;
+
+    > h2 {
+      padding-top: 0;
+    }
+
     &.is-sticky {
       position: sticky;
       top: 0;


braver:
Perhaps the first bucket also needs to be labelled? Now I only understand the first bucket is this year because I run into a 2024 label later. But the first bucket might be last year or even older. Although in some cases ([example](http://localhost:8080/?q=label%3Aanalytics&sort=newest)) I do see 2019 right at the top. Not sure this is intentional, it feels unpredictable.

In one example, I got "June 2016", "March 2016", "December 2015" on one page. I get it, with your explanation of the 3 buckets. But it's also weird in a way. It make me question where May and April with, and all the other month. It suggests the packages between the first and second indicator were all in June, which isn't true probably: they're from any time between March and June. You don't want more than a few of those interruptions, but it's confusing and even a little bit misleading now. 

Maybe instead of simply "June 2016", it should more clearly say "June 2016 starts here", or "June 2016 - March 2016". 

> but I don't think we should do

I agree. Like you say, it's good to get some sense of where things are chronologically, but that's not the main purpose here.

braver:
Sorry, I didn't mean to close! It looked like it had placed a comment I was still writing and then I deleted that comment. I'm not sure what happened.

kaste:
The way I had it in my mind is ... like a timeline but where only a few points have a label and everyone knows that time flies steadily.  So the marker (.timeline-breaks) really just indicate that here the March begins, or here begins 2024.

That being said, it is kinda difficult to grasp.   Sure, just years is likely defensively easy.  But we don't only deal with old vs new data, but with dense and not so dense pages.  If you start searching, you quickly get down to 4 and then maybe 1 result.  At some point we could (a) drop the marker, but (b) also need consistency in some way.  In general, narrowing down is rather quick because we only have 5000 entries in our database.  At 4 entries, it is likely they are on a wide time range.  

Exkurs:
You typically need "gits" http://localhost:8080/?q=gits&sort=update to see GitSavvy  -- everything with year markers
And "gitsa" to *just* get GitSavvy http://localhost:8080/?q=gitsa&sort=update -- marked as September 2025 which is kind of useful as its the date I last updated it.  😕 

II:
selen is enough http://localhost:8080/?q=selen&sort=update -- all three results are marked with the month, if we switch to years after 12 months, the last two entries would not.

All that wasn't even the intent of the change.  I think the intent was to give a clue of where you are in moderately dense pages. 

On just another side, when do you even search in newest mode?  We never had that feature.  We always searched only in in magic ranking mode the last ten years.      

EDIT:  You don't see the initial header, if on page 1 and the heading marker would be the "now" label.  So you might see September 2025 but not October 2025.  

Mhm:  Maybe never show "December 2024" but in that case just "2024".

kaste:
Aside: a sticky search header could be more useful

image

braver:
> All that wasn't even the intent of the change. I think the intent was to give a clue of where you are in moderately dense pages.

Absolutely. All of what I'm saying is like emergent from that. It's a good idea on principle, but then how that plays out in all the infinitely possible search results... is almost impossible to predict.

> You don't see the initial header, if on page 1 and the heading marker would be the "now" label. 

Got it. Although putting "now" up there would be consistent... if superfluous perhaps because you'll figure that out easily enough. 

> when do you even search in newest mode

It's mostly browsing more than searching right? "Let's see what's new in color scheme land" etc. And there it already _almost_ works. I find the labels a bit disorienting, but that can probably be tweaked without abandoning the entire idea. I think it just needs to say "here begins 2024" instead of just "2024", perhaps visually more than textually. 

As a complete alternative I've thought of this, just one indicator above the page explaining the spread over time:

Scherm­afbeelding 2025-10-29 om 10 28 54

... but if the results span a wide period, or if there is only one page, it starts losing meaning ("page 1: 2025 - 2019" ...). But maybe it's an idea we can riff of off.

kaste:
This will take a moment.

kaste:
image

WIP

static/styles.css
```diff
@@ -243,6 +243,69 @@ section {
...
 }
 
+.search-results-header {
+  margin-bottom: 4px;
+
+  > h1,
+  > h2 {
+    margin: 0;
+  }
+}
```
kaste: Ugly shit that I have to overwrite some margins to get it decent.  That is the only variation to the pager right side on the homepage.

static/module/list.js
```diff
@@ -42,21 +60,34 @@ export class List {
   heading = document.querySelector(`[${this.attr}='heading']`)
   list = document.querySelector(`[${this.attr}='list']`)
   hideme = document.querySelectorAll(`[${this.attr}='hideme']`)
+  rangeIndicator = document.querySelector(`[${this.attr}='range']`)
+  pageIndicator = document.querySelector(`[${this.attr}='page']`)
```
kaste: That is a very strange way to query for what are actually sub elements of our component which is the `section`.  I only ever add to this.

kaste: Refactoring will come later

static/styles.css
kaste: Removed

braver:
Awesome, yes this feels like the best balance between the ideas we had. It’s really useful and elegant this way, at least from a few minutes of browsing. Perhaps we’ll have new ideas later, or indeed some tweaks.

Could use some layout tweaking on tiny screens though, it wraps a bit unfortunately:

Search — labellanguage syntax

(Edit: Maybe all it needs is an abbreviation to “p5”)

PR #215, opened 2025-11-03
Fix `margin-bottom` by one pixel

kaste:


static/styles.css
```diff
@@ -257,7 +257,7 @@ section {
 
     @media (width <= 400px) {
       font-size: 22px;
-      margin-bottom: -5px;
+      margin-bottom: -6px;
```
kaste: font-size decreases by 6 pixels and we adjust for that.  🤷

PR #224, opened 2025-11-05
Flip download and release notes links

kaste:
Let the version string link to the package download.  Add a secondary
(optional) link to the release notes.

The order here is clear as we always have a Download link but only
sometimes a Release page.  The "note" icon from octicons is also very
descriptive.

Also slightly simplify requirements text.

### DIFF
--- a/_includes/packages/macros.njk
+++ b/_includes/packages/macros.njk
@@ -14,8 +14,9 @@
 {% macro release(version) %}
   <a class="download" href="{{ version.download_url or version.url }}" title="Download package">{{ version.version }}</a>
 
-  {% if version.sublime_text != "*" %}
-    <span class="requirement" title="Required ST build">({{ version.sublime_text }})</span>
+  {% set requirement = version.sublime_text | format_requirement %}
+  {% if requirement != "*" %}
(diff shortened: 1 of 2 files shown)


michaelblyons:
- I really like how the not-always-present part (the release notes link) is now the supplementary bit of the UI. 👍
- I don't quite understand why you rewrite the >=xxxx to be >xxxx+1. It's not that I disagree... I just don't know why.
- Rewriting xxxx to STxxxx is slightly more informative for people unfamiliar with PC data, but it makes xxxx harder to read.
- ST makes it slightly difficult by having dev releases on a different page from stable, but it could be nice to link the versions to their ref. (Something in this feature list is required for these versions.) This would mean going back to >=xxxx for the first one.

static/style/package.css
```diff
@@ -62,10 +62,9 @@
 
 .requirement {
   color: var(--foreground-2);
-  font-variant: all-petite-caps;
```
michaelblyons: Haha. I was just writing about this.

braver (approved):
:chefs-kiss:

michaelblyons:
> Thanks. We'll touch that code path again I guess.

Probably so. But it does look good. :+1:

PR #227, opened 2025-11-07
Style "More/Other releases" section

### LINKED ISSUE #226: Style for "Other/More releases"
We never styled the Other releases section.  If we want to show also older versions, e.g. for the charts and just for downgrading, we likely need to change that.

kaste:
Fixes #226

The details/summary section to show "Other releases" basically only had the stock design.  Let's change that.

Test with e.g. ["Theme - Monkai"](http://localhost:8080/packages/Theme%20-%20Monokai%20Pro) as that has a lot of "other"/old releases in the workspace.json.

### DIFF
--- a/_includes/packages/macros.njk
+++ b/_includes/packages/macros.njk
@@ -11,11 +11,11 @@
   {%- endif -%}
 {%- endmacro %}
 
-{% macro release(version) %}
+{% macro release(version, omit_requirements=false) %}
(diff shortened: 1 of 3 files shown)


kaste:
Just an iteration; easier if we get more releases in the workspace.  Only then we see our data and can get a feeling for good "reductions".  Likely we have no problems at all if we just show the current (rolling) year.

PR #230, opened 2025-11-14
Implement rank algorithm for "Magic"/relevance

### LINKED ISSUE #229: Magic sort puts a ST2-only package at top-of-list
Searching for PowerShell (https://packages.sublimetext.io/?q=powershell), the magic sort puts PowershellUtils at the top. PSUtils only has a release for ST<3000.

That seems unhelpful, though low impact.

kaste:
Fixes #229

### DIFF
--- a/eleventy.config.mjs
+++ b/eleventy.config.mjs
@@ -96,7 +96,7 @@ function computeMagicMetadata(packages) {
       ? 0
       : clamp01(1 - (ageDaysSinceCreation / MAGIC_LONGEVITY_WINDOW_DAYS))
 
-    const penalty = pkg.removed ? 0.4 : 0
+    const penalty = (pkg.removed || pkg.outdated) ? 0.4 : 0
 
     const baseScore
       = MAGIC_WEIGHTS.popularity * installsScore
@@ -121,6 +121,10 @@ function basePackage(pkg, stat) {
     ...release,
     platforms: util.cleanPlatforms(release.platforms),
   }))
+
+  const supportsModernSublime = releases.some((release) => {
+    return util.parseSublimeTextMin(release.sublime_text) >= 3000
+  })
 
   // For each release, infer a human web URL under key "web"
   // Rules:
@@ -198,6 +202,7 @@ function basePackage(pkg, stat) {
     otherReleases,
     labels: pkg.labels,
     platforms: util.dedupePlatforms(releases),
+    outdated: !supportsModernSublime,
   }
 }
 
--- a/search/index.json.njk
+++ b/search/index.json.njk
@@ -14,6 +14,9 @@ permalink: "/search/index.json"
     "created_at": "{{ pkg.created_at | timestamp }}",
     "last_modified": "{{ pkg.last_modified | timestamp }}",
     "magic_score": {{ pkg.magic_score | default(0) }},
+    {%- if pkg.outdated -%}
+      "outdated": true,
+    {%- endif -%}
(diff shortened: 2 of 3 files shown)


eleventy.config.mjs
```diff
@@ -12,13 +12,120 @@ const supportedRepackagerHosts = [
...
+const MAGIC_FRESHNESS_WINDOW_DAYS = 365 * 2 // bonus for package that had updates
+const MAGIC_LONGEVITY_WINDOW_DAYS = 365 * 10 // Advertise newer packages
+const MAGIC_RECENT_UPDATE_DAYS = 90 // extra bonus for just updates packages
+const MAGIC_WEIGHTS = {
+  popularity: 0.4,
+  stars: 0.3,
+  freshness: 0.2,
+  longevity: 0.1,
+  recency: 0.05,
+}
```
kaste: Here are the magic numbers.  Easy to change.

kaste:
With that in play you get pretty normal  results

image
image

eleventy.config.mjs
kaste: But the weight for "had updates in 3mo" is tiny.  Just a tie-breaker.

eleventy.config.mjs
```diff
@@ -122,6 +122,10 @@ function basePackage(pkg, stat) {
     platforms: util.cleanPlatforms(release.platforms),
   }))
 
+  const supportsModernSublime = releases.some((release) => {
+    return util.parseSublimeTextMin(release.sublime_text) >= 3000
+  })
```
michaelblyons: The wording is slightly misleading, but the effect may be fine. You *can* have a single tag series that has worked since ST2 and still works now.

`parseSublimeTextMin` does what it says, but `supportsModernSublime` doesn't mean the same thing. A `parseSublimeTextMax` could do that.

But maybe it's fine? The penalty is just downranking, and I don't know how many ST2+ packages there are that are still well-maintained, even if they work.

eleventy.config.mjs
```diff
@@ -230,10 +338,13 @@ export default function (eleventyConfig) {
   })
 
   eleventyConfig.addCollection('searchable_packages', () => {
-    return all_packages.map(pkg => ({
+    const searchable = all_packages.map(pkg => ({
       description: pkg.description,
+      installs_window: stats[pkg.name]?.installs?.yearly?.reduce((a, b) => a + b) ?? 0,
```
kaste: ☝️ That is the installs window

kaste:
Damn.  That looks like a bug.  My first thought was, well their plugins have dependabot cadence for the updates, and that might downvote the main package, but LSP had a recent update, likely the highest installs, likely the most stars.

michaelblyons:
> but LSP had a recent update, likely the highest installs, likely the most stars.

Yes, Yes, and Yes. :confused:

kaste:
LSP's Minisearch(!) score is only 0.475, LSP-eslint (and most other in the top have) ~0.9.

michaelblyons:
Boost the name and mega-boost if name exact-matches?

kaste:
We need to check if having a token twice or trice in the description changes the score.  That would mean that we don't build a set of words from the input -- in other words we don't deduplicate the input. That is natural for long texts but for a subheading/description it seems like a mistake.  

Generally, to rank with the text score seems too vague as we don't have that much text indexed, it is always just a name and a description.  (Google would index the README, obviously.)

PR #232, opened 2025-11-14
Mark outdated, st2-only and st3-only, packages

kaste:
We create two synthetic labels "st3" and "<st3", style them a bit different and get a lot for free.  

image

### DIFF
--- a/_includes/packages/macros.njk
+++ b/_includes/packages/macros.njk
@@ -60,7 +60,7 @@
               {{ label }}
             </a>
           {% else %}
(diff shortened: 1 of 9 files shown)


michaelblyons:
I agree with the top two posts, though I think the top-right labeling should use the "archive box" of R.I.P., although maybe a less intense color.

Seems reasonable to ship for now, though.

kaste:
Actually, one could argue to just invent synthetic RIP and MIA labels as well, then we can search for them and have a very clean top-right corner.  🤔

Issue #82, opened 2025-08-15
Some labels should stand out...


kaste:
Initially, I thought just theme and syntax colors should stand out.  Sparkle?  

<img width="1408" height="1018" alt="Image" src="https://github.com/user-attachments/assets/7ec7a0c2-959a-4e6b-ae89-d5acbcc78b71" />

Then I also added the actual syntax defs:  

<img width="1428" height="1042" alt="Image" src="https://github.com/user-attachments/assets/1b7108b9-a066-44ac-855b-14ab95044447" />

braver:
Maybe all 6 that were picked to get highlighted below the search field should get some kind of sparkle ✨

kaste:
That wasn't the idea.  The idea was not to color code something or to recreate the Christmas tree.  Mostly everything is monochrome and then we add some sparkle.  Even coloring just lang syntax makes it inharmonic as we had too much coloring ob the whole page.

I identified theme plus scheme because they literally add color to our editor.  (They are also very popular in general, and basically things people like to try out; there is no buy-in cost for them.)  The gradient comes from the observation that packages often have both labels, and then the gradients -- because they have a direction -- were meant to visually connect the two labels.  

Language tags maybe deserve their trademark icon, usually the same as their file extension icon.   So Python would get the Python icon in front etc.  

We can also just highlight the labels if we filter by them or if we hover of them in the search form....

That being said, just removing the platform colors looks like that

<img width="1327" height="1042" alt="Image" src="https://github.com/user-attachments/assets/f628c2a3-b8e7-4d55-8bb7-d1b77e807330" />

Maybe that's sad, or minimalistic, or we just need a moment to adjust to that.

kaste:
Hm... 

<img width="893" height="246" alt="Image" src="https://github.com/user-attachments/assets/75def4f4-41e0-4594-89ba-00dd0bea2c38" />

michaelblyons:
I can't deny that's cute. 🙂

You may want to drop some of the languages if this ships. (Or may not. 🤷) Tags like `diff` or `todo` might indicate a language, but they might just indicate functionality. And you just make a choice to use or not use the icon for each one. The `ai` icon in AFileIcon is for Adobe Illustrator, but we all know [what that tag means, now](https://packages.sublimetext.io/?q=label%3A%22ai%22). 😅

FichteFoll:
[...]

PR #236, opened 2025-11-19
Synthetic labels for RIP, MIA, etc

### LINKED ISSUE #234: Invent synthetic RIP and MIA labels
One could argue to just invent synthetic RIP and MIA labels, then we can search for them and have a very clean top-right corner.  🤔

_Originally posted by @kaste in https://github.com/packagecontrol/thecrawl/issues/232#issuecomment-3534969385_
            
What to like:  You can click on such a label and the search will show all packages with that label.  Clean visual design, a bit less clean from the programming/domain model.  Likely manageable.  You could as well have "labels", a copy of the labels in the db, and collect "synthetic_labels", also a `list[str]`, and then present them in the same visual spot.  

We also put the platforms here visually, kind of synthetic labels as well.

braver:
Fixes #234 

Pushing what I have so far to get feedback. I just have the visuals and the markup, I didn't actually implement search/filtering (yet): clicking e.g. "unmaintained" will give no results. I intend to implement that once I remember how that stuff works 😅 

- moved the RIP/unmaintained labels out of the top left
- they now join "st3" etc together with the labels and platforms in the bottom left (looks very clean IMO), in lower-case
- I changed "<st3" to "st2" because that's what [pc.io does](https://packagecontrol.io/packages/Clipboard%20History) and it's easier to read (I think "st1" never had package control? it's very long ago) - let me know if that's a mistake
- I propose to not ride along on the `labels:"foo"` filtering because these are not actually labels. It's certainly easier to implement, but I think it muddies the data you're actually filtering on. Instead perhaps introduce `state:"foo"` as follows:
  - state:rip (removed)
  - state:unmaintained (archived_at)
  - state:st2 (outdated)
  - state:st3 (st3_only)

Example card:
Scherm­afbeelding 2025-11-19 om 15 40 33

Example detail page:
Scherm­afbeelding 2025-11-19 om 15 40 40

Another card:
Scherm­afbeelding 2025-11-19 om 15 41 10

Last one for good measure:
Scherm­afbeelding 2025-11-19 om 15 53 18

### DIFF
--- a/static/style/package.css
+++ b/static/style/package.css
@@ -130,12 +130,19 @@ ul.stats {
     padding: 2px 6px;
     border: 1px solid transparent;
     &.state {
-      border-color: var(--orange-1);
-      color: var(--orange-1);
+      background: inherit;
+      border-color: var(--orange-2);
+      color: var(--orange-2);
+      &:hover {
+        background: rgb(from var(--orange-1) r g b / 0.12);
+      }
     }
     &.state-bad {
-      border-color: var(--red-1);
-      color: var(--red-1);
+      border-color: var(--red-2);
+      color: var(--red-2);
+      &:hover {
+        background: rgb(from var(--red-1) r g b / 0.12);
+      }
     }
     &.platform {
       background: var(--blue-1);


michaelblyons:
- FWIW, I think `r.i.p.` looks strange. I think moving to `archived` or `removed` is better than changing `unmaintained` to `MIA`. (Though `archived` as a tag name kind of conflicts with the fact that `unmaintained` means the GitHub repo is archived.)

- I like the ideas both to add to the labels page and to add the crawler failures as a distinct option.
    - Minor point: If crawler failures are added, I hope that we can use the same verb form as the other states, like `crawl-failed`.

- Styles in OP are fine with me. Other styles might also be fine.

- No opinion on removing the platform labels from the front page.
    - I have occasionally thought that using the same style for **all** the platforms might be okay, especially if #82 goes forward.

> You don't get me if you don't show the meat, the easy way to search.

Fair. I do it's worth considering the UX too, even if the functionality hasn't been written.

> the backend needs a feature split here.

This part I'm not sure I understand. Do you mean `label:X` vs `state:X`/`platform:X`/`label:X`?

static/module/card.js
```diff
@@ -124,11 +103,33 @@ export class Card {
     }
 
     this.pkg.labels.split(',').forEach((item) => {
-      parent.appendChild(this.button(item))
+      console.log(item)
```
kaste: Yo.  Seen that 🙃

eleventy.config.mjs
```diff
@@ -191,8 +191,14 @@ function basePackage(pkg, stat) {
...
   const labels = pkg.labels?.slice() ?? []
-  if (!supportsModernSublime) labels.push('<ST3')
+  if (!supportsModernSublime) labels.push('ST2')
   if (supportsModernSublime && doesNotSupportNewestSublime) labels.push('ST3')
+  if (pkg.archived_at) {
+    labels.push('MIA')
+  }
+  else if (pkg.doa || pkg.removed) {
+    labels.push('RIP')
+  }
```
kaste: That's the wrong precedence.  We don't show MIA if already RIP.

braver:
> On the card, I would do e.g.

Love that, looks good 👍🏻 

> Btw "doa" isn't a thing anymore as I "backported" first_seen for all packages

Oh, that's certainly good to know. If I can drop DOA that simplifies this PR a bit as well.

Issue #242, opened 2025-11-23
Labels search: explain why results count is off more clearly


braver:
It says it's omitting items with one usage, but I would expect them to be omitted completely and not counted. The number of results doesn't match the number of items I'm seeing, making me do a double-take (and admittedly I first thought there might be a bug). 
Maybe we can say "0 results (1 omitted)"?

<img width="946" height="564" alt="Image" src="https://github.com/user-attachments/assets/024cc968-c6a1-4529-9636-737c81a2054a" />

kaste:
I think that's a fake one.  You already played with the "omitting" button.  In pristine state, searching for 3d would have resulted in 

<img width="920" height="423" alt="Image" src="https://github.com/user-attachments/assets/c9f87142-a668-49f1-9e1e-0501588ae5e6" />

But when you play with the omit button, the guess is that you learn what it does by using it.   

Likely I started with: 0 results, matching what we see, but that is fake, there is one result, it is just hidden.  (Omitting's synonym in this context is "not showing".)  Then we have "1 result" which is fake because it is showing none, and then you show "1 result, 1 omitted" which is actually already on the screen, as it says 1 result   ------ Omitting (Not Showing) labels with rare usage.

braver:
Perhaps a PR better explains what I had in mind: #243

braver:
> I usually think "Results" == matched the query. And "Omitting" is results I'm not showing

I get what you’re saying. I think my point is that the UI in its current state doesn’t necessarily communicate that the omissions overlay the search query results. Its position next to the search field makes it feel like a filter (and part of the query). You can argue that’s it’s not, and that I can discover that it’s not, but I think it can be made to work more like a filter and fit that expectation.

> a puzzle you figure out basically emotes joy

That’s not actually true in all cases. A UI should typically not aim to be a puzzle, and solving it can be frustrating more than joyful.

Semantics aside… 

My point is not that the current implementation is wrong. Just that it has potential to be subtly improved. And the wording can be changed to avoid double negatives (double negatives and UX are worth a read if you’re interested). Hence my PR.

braver:
Also, to be clear, I just wanted to put my ideas out there, in text and actual code, and to discuss and explore the topic. I’m not sure or convinced one way or the other, opinions weakly held here for sure.

kaste:
> A UI should typically not aim to be a puzzle

[...]

PR #243, opened 2025-11-23
Tweaks for label search and singles

braver:
#242 - Explain how many items are not visible due to omitting singles

You can get into a situation where the count of results doesn't match the number of items you're seeing. If you remember interacting with the "omission toggle" you can find the explanation yourself, but a good UI should explain itself. Of course, as discussed in the linked issue, the number of items is somewhat irrelevant... but it's there and I find it interesting data (e.g. how many are actually single-use labels?).

I reworked this PR to be minimally invasive in terms of look and feel compared to earlier attempts. I merely moved the toggle to be more part of the list and less of the filters, where other meta-data of the results is also typically displayed. And added a counter for how many items are hidden.

https://github.com/user-attachments/assets/bdfa80e0-a66b-443e-85dd-942317dd0c74

### DIFF
--- a/static/style/search.css
+++ b/static/style/search.css
@@ -152,7 +152,9 @@ select {
     display: flex;
     align-items: center;
     gap: 0.5rem;
-    padding: 0 0.75rem 0 1rem;
+    padding: 0 0.75rem 0 1.5rem;
+    margin-left: 0.5rem;
+    border-left: 1px solid var(--background-4);
 
     button {
       background: none;


FichteFoll:
"20 of 108 Results" reads like pagination even though the page is not paginated. I much prefer the initial version.

braver:
Pushed. It's all in the details ;) PR state on top:

Scherm­afbeelding 2025-12-07 om 21 08 43

kaste:
Although I did like portions of the UI over the at this point deployed one; it was at the same time to disappointing sloppy for a rev3 variant.  E.g. the focus-visible just outlined half the header.  For the normal interaction, pristine state and search for something, the UI went from "Omitting 567 labels" to "~~Omitting 567 labels~~" to "~~Omitting labels~~".  Noisy and full of information nobody ever asked for.

PR #250, opened 2025-11-25
Implement a way to re-enter auto mode on the labels page

kaste:
Mitigates #242 

By re-entering auto-mode, new users that just clicked around will less likely see a less useful state.


braver:
Makes sense 👌🏻

PR #251, opened 2025-11-25
Show language icons on the labels

kaste:
Ref #82

### DIFF
--- a/util/build-label-icons.mjs
+++ b/util/build-label-icons.mjs
@@ -5,7 +5,7 @@ import path from 'path'
  * Build an SVG sprite and JSON mapping for label icons based on
  * SVGs from repo at `.AFileIcon`.
  *
- * Input:
+ * Assumed Input:
  *   .AFileIcon/icons/svg/file_type_<label>.svg
  *   .AFileIcon/icons/icons.json
  *


braver:
Hey, that's really really cool actually. 

Typically all we have is monochrome svg, for these things, as that's also what Atom, and then VS Code had. So these collections never retained the full color... It is very clean and readable at small sizes this way, so it's not all bad.

michaelblyons:
> these icons are mostly monochrome

Is the trouble that each icon only uses a single color or that all of them use the *same* single color? i.e. What if Docker was blue?

braver:
There should be a mapping of icons to colors. I use the same approach for [FileIcons](https://github.com/braver/FileIcons/blob/master/build/icons.json) (because it's based on A File Icon but since deviated completely). The SVG will have no predetermined color, or just black. You can inject the colors into the svg in a build step, or as we do here since ST only supports PNG generate hi-res PNG from the svg with the color injected. If you have a classname we can also overlay the color in CSS, but that's slightly less elegant IMO.

braver:
```css
  --label-icon-color-red: var(--red-2);
  --label-icon-color-orange: var(--orange-2);
  --label-icon-color-yellow: var(--orange-2);
  --label-icon-color-green: var(--green-2);
  --label-icon-color-cyan: var(--blue-2);
  --label-icon-color-blue: var(--blue-2);
  --label-icon-color-purple: hsl(275 76% 58%);
  --label-icon-color-pink: hsl(345 76% 58%);
  --label-icon-color-sky: var(--blue-2);
  --label-icon-color-graphite: hsl(245 21% 42%);
  ```
  
All those blues are very close, and the lighter ones lack readability. Same for yellow, most non-ugly shades are invisible. This does add a pink and a purple, and a purplish "graphite", for some variation.
  
Scherm­afbeelding 2025-11-26 om 18 35 19

kaste:
I didn't do a search before but actually there is also https://github.com/devicons/devicon/tree/master/icons.  Just as easy to extract. Their strangely green website: https://devicon.dev/

Your colors are good.   Same for dark mode?

kaste:
I didn't look further into my code, but generally I tend to commit the information we extract from AFileIcon even when it is "easy" to recreate.  I don't want any breakage in the middle of the night in the next 10 years.  AFileIcon is a dependency we don't control after all, and just renaming a file/folder would break us.  What one can do: make a CI action, check for changes, and then open a PR or write an email so we can merge the changes.  (Merely to get new icons.)  

What is missing:  a filter in the config so we can skip some icons we don't want.

kaste:
>  looked a lot better on my machine

Every color?

michaelblyons:
> I don't want any breakage in the middle of the night in the next 10 years.

Make it a submodule, so you get the exact commit you expect. Then if you want, Dependabot can PR you with updates, and you don't even have to keep the SVG or JSON content locally.

kaste:
*Personally* I don't want to maintain this as well;  I don't want to depend on Dependabot and I don't want to get spammed by PRs "randomly" during the course of a year or even years.  Even ignoring notifications is distraction, 😀.

PR #256, opened 2025-12-03
Include "labels" when doing a free form search

### LINKED ISSUE #235: Check if having a token twice or trice in the description changes the Minisearch score
Check if having a token twice or trice in the description changes the score.  That would mean that we don't build a *set* of words from the input -- in other words we don't deduplicate the input. That is natural for long texts but for a subheading/description it seems like a mistake.  

Generally, to rank with the text score seems vague as we don't have that much text indexed, it is always just a name and a description.  (Google would index the README, obviously.)

_Originally posted by @kaste in https://github.com/packagecontrol/thecrawl/issues/230#issuecomment-3534992825_
            
As soon as we have the ranking implemented, we can use Minisearch merely to *find* packages and less so to also do the sort-by-relevance as the "relevance" is *not* in the literal text fields.  (I guess.)

kaste:
This in effect requires a new ranking algorithm which is then in the second commit and closes #235.

### DIFF
--- a/static/module/sort.js
+++ b/static/module/sort.js
@@ -91,25 +91,34 @@ export class Sort {
   }
 }
 
-const MINISEARCH_SCORE_WEIGHT = 0.2
+function shiftedSCurve(x, a = 2, b = 0.5) {
+  const u = x - b
+  const num = Math.pow(u, a)
+  const den = num + Math.pow(1 - u, a)
+  return num / den
+/*
+def shifted_s(x, a=2, b=0.5):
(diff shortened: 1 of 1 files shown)


kaste:
In the simples form, searching "nss" did not find the new package

image

image

kaste:
We could of course discuss if complex search queries like `label:web component` still search for component in labels.  (Because "label" is explicitly asked for in the query.)  But we don't do that for author as well, so that will rather be a future addition.

Issue #260, opened 2025-12-05
Magic sort still has weighting problems


michaelblyons:
See the list of language-syntax packages: https://packages.sublimetext.io/?q=label%3A%22language+syntax%22

I'm very suspicious of Sass and Git not being on the front page

michaelblyons:
- Sass has 1.7M downloads, 51 stars (which is admittedly low), and was updated in the last 2 years.
- Somehow, YamlPipelines, at 1.4k downloads and 24 stars---but was updated last month---manages to outrank Sass.

FWIW, I have YamlPipelines installed and not Sass (not anymore, though I used to a few years ago), so maybe the magic is giving good suggestions for *me.*

As another example, on [my list of authored packages](https://packages.sublimetext.io/?q=author%3A%22michaelblyons%22), it seems particularly odd to see [Crontab](https://packages.sublimetext.io/packages/Crontab) ahead of [Hosts](https://packages.sublimetext.io/packages/Hosts), which I think should outrank it on every single criteria...?

A last example, where I think the results are _good,_ just surprising: Searching for "docker" ranks Containerfile well above Dockerfile Syntax Highlighting. I think Containerfile is objectively better, but it loses by an order of magnitude on every criteria except recent updates.

TLDR: The results do seem both "good" and also "magic," considering that I thought I understood the ranking algm and I'm still perplexed.

kaste:
I don't think the MB downloads count;  didn't we switch to recent downloads for the ranking?  You switched from Git to YamlPipelines in the reply.

The language syntax thing is anyway confusing, you would expect Python3 improved, SuperTypescript. Idk a new language that doesn't ship a syntax already.  I don't think phpfmt even ships a syntax, and PlainTasks only ships a syntax for the dashboard views.  That's as if GitSavvy would ship language syntaxes where it just defines some for its own proprietary special-views.  IMO same for [rainbow_csv](https://packages.sublimetext.io/packages/rainbow_csv), that's just using a syntax for colorizing a very limited format.

kaste:
[Dockerfile Syntax Highlighting](https://packages.sublimetext.io/packages/Dockerfile%20Syntax%20Highlighting) is missing the docker label.

kaste:
If we had an UI idea I could also print all the scores somewhere.

michaelblyons:
You could throw them in an HTML comment even without a UI.

PR #262, opened 2025-12-07
Print detailed ranking insights in comments above the card

kaste:
Ref #260 

The patch is a bit bigger; if we revert at some point the changes in search.js and sort.js are still good to keep IMO.


kaste:
Nerdy, in a yes we can style

image

PR #264, opened 2025-12-09
Disable omit toggle while searching

### LINKED ISSUE #242: Labels search: explain why results count is off more clearly

kaste:
Fixes #242
Closes #243 

This effectively enforces the `autoMode` and disallows toggling the omit
state while in a search.

Since we don't do anything special when showing search results, the
explaining sentence "Omitting ..." is slowly faded away.

The implementation reduces the state management to two attributes on
the element, namely "disabled" and "aria-pressed".  The elements state
is thus the source of truth.

### DIFF
--- a/labels.njk
+++ b/labels.njk
@@ -24,9 +24,16 @@ date: Last Modified
   </div>
 
   <div class="search-subcontrols">
-    <button type="button" class="usage-toggle" aria-pressed="true">
+    <button type="button" class="usage-toggle" data-omit-control aria-pressed="true">
       Omitting labels with only one usage
     </button>
+    <script>
+      {
+        if (document.documentElement.classList.contains('initializing')) {
+          document.querySelector('[data-omit-control]').setAttribute('disabled', '')
+        }
+      }
+    </script>
   </div>
 </form>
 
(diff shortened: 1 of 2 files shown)


static/style/search.css
```diff
@@ -306,10 +306,16 @@ form.compact {
...
-    &.is-active {
+    &[aria-pressed='true'] {
       color: var(--foreground-1);
       font-weight: 500;
       text-decoration: none;
     }
+
+    &[disabled] {
+      cursor: default;
+      pointer-events: none;
```
braver: TIL about [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert), which is a bit more comprehensive than just `pointer-events`. This is fine though, the combination of `:disabled` and removing `pointer-events` does about the same. Just wanted to share that :)

braver (approved):

PR #268, opened 2025-12-12
Let the result counter match the actual result set

### LINKED ISSUE #242: Labels search: explain why results count is off more clearly

kaste:
Fixes #242


kaste:
Don't overthink and KISS

PR #271, opened 2025-12-19
Bump to a 28px grid

kaste:


kaste:
I added comments here already  https://github.com/packagecontrol/thecrawl/commit/3b4ab7f4565934deca76974d2e0d814e50d53720

braver:
> You: We should go defensively here. .... Changes 7 files.

I never said we should go defensively. Sorry if that got confused. My point in the original issue was that changing the rhythm of the typography upsets everything, and that it would take changing 7 files to make everything sit correctly again. Hence this PR. Even going from 26 to 28 upsets the balance. There is a logic to how things fall into place across the entire design.

It's not a bad thing, it just needs to be followed through on. I see your points about the readme's specifically. I'll have another look at that.

Also look at the cards by the way, I do feel they are improved by this PR.

kaste:
The cards go from 
image
to 
image

That's very subtle more dense.  The heading of the package details page is better and a bit less subtle so.  

I don't think the About page has improved as it was perfect before IMO (e.g. with the generous WS).

braver:
> That's very subtle more dense

There is a subtle change in the alignment of descriptions of the cards. It's all very subtle.

kaste:
Yeah, interesting.  What I do like about Github is how easy they manage the spacing between the paragraphs and headings.  It is always the same margin-bottom (16) and except for p the same margin-top (24).   That is very uniform.

I think we loose the density or conformity to a rhythm in exactly these h-p spacings.  The minimal patch with the most impact is maybe still 

```
.readme {
    margin-top: 3.2rem;
    padding: calc(var(--side-padding) * 1.7) calc(var(--side-padding) * 4.3);   /* <---- */
    border-radius: 6px;
    background: var(--background-1);
}
p, ul, ol, summary {
    font-size: 1rem;
    line-height: 28px;   /* <-- */
    padding-top: 0px;  /* <-- */
    margin-bottom: 16px;  /* <-- 1rem */
    padding-bottom: 0;
    margin-top: 0;
} 
code, kbd {
    font-family: "IBM Plex Mono", monospace;
    font-size: 0.94em;  
    line-height: 1rem;  /* <-- */
}
```
which gives
image


Looks like more uniform spacing for the h3's and the p's.  Basically, we don't loose space, density just because of the headings.  And still have the comfort of 28px line-height.  

Always important, what we not discussed yet, the line-length was too big -- the padding change to x4.3 in the patch above.  Less words per line, better readability.  

So, I don't think we need the information density of Github.  28px p line-height, less line-length, and a KISS 1rem margin prove it.

braver:
> Always important, what we not discussed yet, the line-length was too big -- the padding change to x4.3 in the patch above. Less words per line, better readability.

Lines are definitely too long in the readme. But what about the left-over space in the readme block? Let it dangle? Or do we simply make the readme block less wide. I had it less wide before (ages ago), we made it wider ... no simple answers for any of these things. I'll try a few things.

> So, I don't think we need the information density of Github. 28px p line-height, less line-length, and a KISS 1rem margin prove it.

Thanks for the patch, I'll give that a go.

kaste:
> dangling ws 

The association is the piece of paper on a page.  I had roughly 1100px reading width.

image
image

kaste:
README on Github is like this

image

Roughly >1000px.

Please add 

```
  table {
    border-spacing: 0px;
    border-collapse: collapse;
    margin-bottom: 1rem;

    tr:nth-child(2n) {
      background: rgb(from var(--background-2) r g b / 0.35);
    }
    
    td {
      border: 1px solid var(--background-3);
      padding: 6px 13px;
    }
  }
```
 to the mix.

800px (52rem) can be not enough for content you don't control; and we have to proof read it against 100 example packages..

braver:
Some resources:

> Line length should be an average of 45–90 characters per line
> - https://practicaltypography.com/typography-in-ten-minutes.html 

> keeping lines under 80 characters for non-CJK languages
> https://www.uxpin.com/studio/blog/optimal-line-length-for-readability/

Now 1 rem is not 1 character wide. But somewhere between 40 and 60 is typically what you're looking for.

> Do it! 52rem in the middle, like paper. 👍

Yes, looks really good 👍🏻

braver:
So, this got kinda long and detailed and touches a lot of css. But that's what happens when you start shifting and nudging things around. Good stuff though:
- better spacing around headings and lists
- better tables, inline code and code blocks
- better line height and length especially in readmes
- some more attention to details across the board from the header to the footer

Especially readme's are much improved. I may have missed something... there is always space for improvement. Overall though, right now, looks good to me. Might be a good time to merge and save more iterations for later.

kaste:
image

braver:
Well damn, I spent like an hour tweaking the pager thing and I think it looked fine on my machine. But I did forget about the status page. I'll also look at increasing the density there 👁️

kaste:
I'm actually feeling bad for this critique after two weeks.   💅

PR #273, opened 2025-12-19
Add simple status page for the crawl

kaste:
thecrawl already logs at the releases page but that has no history.  Time to pump it up.

image

With nifty keyboard navigation.
http://localhost:8080/status


michaelblyons:
You could link the word "Built" in footer to this page. Some time ago, I thought about linking that to the workflow runs on GH itself, but this is even better.

PR #300, opened 2026-01-10
Common data model

deathaxe:
A model is something shared. It describes a common interface or protocol all participants have to adhere to.

This commit therefore consolidates code hosters' repo related data type declarations in a common module to ensure all drivers use compatible interfaces and produce same output for consumption by crawler.

### DIFF
--- a/scripts/bitbucket.py
+++ b/scripts/bitbucket.py
@@ -5,47 +5,15 @@ import asyncio
 import os
 from urllib.parse import urlparse
 
-from typing import AsyncIterable, TypedDict, Literal, Iterable
+from typing import AsyncIterable, Iterable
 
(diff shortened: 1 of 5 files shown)


kaste:
That doesn't add much value as it fakes a "common" model which we actually don't have.  It lacks the modelling so to speak as it basically is a copy-paste to a shared file.   

E.g.  the so-called `RepoInfo` now lacks a "releases" member GitHub actually has and the others not.  A sum type instead of a common type seems more appropriate.  

E.g.  the metadata object which we just grab into the db (`out = info["metadata"] | out` );  we grab whatever we get.  If you want, we implicitly expect that it is serializable but besides that we don't expect a special form I guess.

PR #309, opened 2026-01-18
Simplify platform labels

### LINKED ISSUE #283: Simplify platform labels
I still would like to see a change how we present the platform buttons/labels.  Rationale: 99% of packages are cross-platform; and if not, it is easier to grasp: this package does *not* run on Windows. Examples always compared to status-quo:

Image

Image

Image

Image

Image

Image

However, the styling of the strike through / cross out is not as simple as I would have thought.

Image

Image

 
Image

Image

Image

Surprisingly extra hard to decipher:

Image

Image

braver:
Fixes #283 

This is a first pass at what we discussed in said topic.

Scherm­afbeelding 2026-01-18 om 16 35 26
Scherm­afbeelding 2026-01-18 om 16 35 00
Scherm­afbeelding 2026-01-18 om 16 34 55

Clickable:
Scherm­afbeelding 2026-01-18 om 16 35 16

I did not change anything about libraries (actually haven't even looked at that page yet). Not sure if that should be in scope or not?

### DIFF
--- a/eleventy.util.mjs
+++ b/eleventy.util.mjs
@@ -55,6 +55,9 @@ export function computePlatformLabelsForSearch(releases) {
 export function computePlatformStatement(platforms) {
   if (!platforms.length) return ''
   if (platforms.includes('any')) return ''
+  if (platforms.some(token => token === 'osx' || token.startsWith('osx-'))) {
+    throw new Error(`computePlatformStatement received non-canonicalized platform: ${platforms.join(', ')}`)
+  }
 
   const tokens = []
   const variantsByOs = new Map()


kaste:
Yeah.  From the internal model I would have gone the way to define a `platformStatement`
in the model, in eleventy.blah.mjs, so that we can have tests for this and have
as many variations as we need them.  (E.g. only for arm, only for x32)  I would
have guessed this to be future-proof and simple to maintain.

Then we just display this statement wherever it fits and if there is any.

I would not make this clickable.  Esp. not when it is displayed in areas where nothing 
else is clickable.  (We can still make it clickable later if we have a good href/search.)

I esp. wanted to get rid of "Find more packages for MacOS" as that is low-value.
(Compared to "Find similar packages for Windows/Linux".  That would be 🔥.)

The placement on the search cards seems obvious, and there is lots of space here.
The placement on the small cards (on the homepage) seems too dense; the vertical 
space between the right-top lines and the label-buttons is irritating.  

I would play with putting them (a) on the same line with the buttons, (a1) one left, the other right, (a2) just on a line;  (b) if you left align the buttons, the top-right spacing
is likely a no-brainer, but from my experience it looks shit for the normal case
where we don't have such a statement in Desktop mode.  On smaller screens, iirc
when we first have 2 then 1 column, it looks good left aligned in all cases; but for 
the normal Desktop mode with three columns it did not.

On the details page, I would have chosen the same placement, now that I see it maybe 
use the middle point we use elsewhere already because we don't have an icon on the left of 
the statement.  

So everything al'right except the homepage cards.

braver:
> I would not make this clickable.

I agree. I built it without first, then added it only because it was easy. But indeed visually it's not really discoverable or logical, and it doesn't provide value.

> The placement on the small cards (on the homepage) seems too dense

I considered removing them there because it doesn't look right... but that's not ideal.

> I would play with putting ... etc

I'm going to have to come back and read that a few times before I get al that 😅 
Anyway, in these cards there simply is no space left, unless more new space is created, or we combine things more efficiently. Example of a densely filled one: 
Scherm­afbeelding 2026-01-18 om 21 09 06
Maybe downloads and stars on one line, and accept that sometimes forces a package name to break to the next line. And then the "not for platform" on the second line.
... I can play with that.

> On the details page

There it just fell in place that way almost by itself, so I left it untouched basically. 

So I can definitely do a pass at wrapping up look and feel.

As for ...

> From the internal model I would have gone the way to define a platformStatement
in the model, in eleventy.blah.mjs, so that we can have tests for this and have
as many variations as we need them. 

... I just had like half an hour to waste this Sunday and half a hour over a week ago. I simply don't have the time to do it properly and refactor around this. Or at least not quickly. I also really don't mind if you take this and run with it, perhaps after I improve the home cards over the next few days.

static/style/grid.css
```diff
@@ -104,8 +104,10 @@ section[name="labels"] .grid {
     padding: 0;
     margin: 0;
     line-height: 1.6;
+    line-height: 18px;
```
braver:
```suggestion
    line-height: 18px;
```

static/style/grid.css
```diff
@@ -126,6 +128,19 @@ section[name="labels"] .grid {
...
       padding: 1px 6px;
     }
+    .platform-statement-wrap {
+      flex: 1 1 auto;
+    }
+    .platform-statement {
+      background: transparent;
+      margin: 0;
+      padding: 1px 0;
+      text-underline-offset: 2px;
```
braver: I don't believe we actually underline this text?

static/style/grid.css
```diff
@@ -126,6 +128,19 @@ section[name="labels"] .grid {
       line-height: 20px;
       padding: 1px 6px;
     }
+    .platform-statement-wrap {
+      flex: 1 1 auto;
+    }
+    .platform-statement {
+      background: transparent;
+      margin: 0;
+      padding: 1px 0;
```
braver:
```suggestion
      padding: 0;
```

Since the text here is base-line aligned, this padding should not affect how the statements sits compared to the other labels. The 1px top/bottom padding seems to add a bit of unnecessary space. This is with zero:

Image

eleventy.util.mjs
```diff
@@ -17,14 +20,141 @@ export function cleanPlatforms(platforms) {
 export function dedupePlatforms(releases) {
   const all = releases.flatMap(release => release.platforms)
   const unique = Array.from(new Set(all))
+  if (unique.includes('*')) {
+    return ['any']
+  }
   // when a package actually supports all platforms (across multiple releases)
-  if (unique.includes('linux') && unique.includes('windows') && unique.includes('macos')) {
+  if (unique.includes('linux') && unique.includes('windows') && unique.includes('osx')) {
```
braver: I believe a side effect of this is now you have to search for `platform:osx` to find packages for macOS, while before you would search for `platform:macos`. If you know the underlying data, osx actually makes sense... perhaps we should support both?

kaste: Good catch, I reverted to the old behavior.  To have both must be a separate issue.

static/style/grid.css
braver: I know, and it's fine, but I do sometimes get lost in my own muck and then I'm like "come one, you do this professionally, you should be able to do figure this out" 😄

PR #325, opened 2026-01-27
Add GitHub style markdown alerts

### LINKED ISSUE #296: Support for Github Alerts in README's
How about rendering Github Alerts in READMEs?

Some packages make use of them, and they look pretty bare (or broken) on packages.sublimetext.io

Examples:

```markdown
> [!NOTE]
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]
> Crucial information necessary for users to succeed.

> [!WARNING]
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

Possible extension: https://github.com/bent10/marked-extensions/tree/main/packages/alert

kaste:
Fixes #296 

Also adds header anchors for deep linking into the readme.

image

### DIFF
--- a/static/style/readme.css
+++ b/static/style/readme.css
@@ -37,25 +37,11 @@
 
   code {
     background: var(--highlight);
-    background: hsl(208.12deg 80% 92.16%);
-    :root[data-theme='dark'] & {
-      background: hsl(212.96deg 29.41% 23.46%);
-      background: hsl(212.96deg 34.44% 29.83%);
-    }
   }
 
   pre {
     border-radius: 6px;
     background: var(--highlight);
-    background: hsl(36.74deg 100% 94.71%);
-    background: hsl(36.74deg 100% 97.06%);
-    :root[data-theme='dark'] & {
-      background: hsl(9.05deg 16.96% 23.44%);
-      background: hsl(231.22deg 13.01% 25.35%);
-      code {
-        background: transparent;
-      }
-    }
   }
   pre.fallback {
     background: var(--background-1);
@@ -72,10 +58,6 @@
     padding: 0px 4px;
     box-shadow: inset 0px -1px 0px var(--background-1);
     background: var(--highlight);
-    background: hsl(208deg 100% 97.06%);
-    :root[data-theme='dark'] & {
-      background: hsl(208deg 31.75% 24.45%);
-    }
   }
 
   table {
@@ -94,15 +76,18 @@
   }
 
   a {
-    color: var(--foreground-1);
-    text-decoration-color: var(--foreground-4);
-    text-underline-offset: 4px;
+    color: hsl(32deg 5.26% 28.36%);
+    text-decoration-color: hsl(207.91deg 71.14% 47.81%);
+    text-underline-offset: 3px;
     transition: color .1s ease-in-out, outline .1s ease-in-out, text-decoration-color .1s ease-in-out;
 
     :root[data-theme='dark'] & {
-      text-decoration-color: var(--foreground-2);
+      color: hsl(32deg 20.52% 81.99%);
+      text-decoration-color: hsl(210.05deg 74.64% 53.69%);
+      text-underline-offset: 3px;
 
       &:hover {
+        color: var(--orange-2);
         text-decoration-color: var(--orange-2);
       }
     }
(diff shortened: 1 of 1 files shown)


static/style/readme.css
```diff
@@ -74,4 +74,70 @@
       padding: 6px 13px;
     }
   }
+
+  a {
+    color: hsl(31.78deg 88.74% 37.93%);
```
braver: This kind of goes back to our earlier discussion about color contrast: it's the wrong kind of orange. It's not our "brand" color. And because we already set the expectation with a certain kind of orange everywhere else, these links are now obviously different somehow. My first reaction was that maybe they were links I already visited ... it looks wrong to my eye.

I don't mind the underlines by the way, more links should have underlines for various reasons.

braver:
> The heading anchors are traditional links but don't have text in it and no aria-* labels; so that's already not so super good; you can tab them but the outline is ugly 🤷

Since the icon is revealed on focus, perhaps the outline is not needed? Alternatively, looks better if you make the element square and center the icon:

Scherm­afbeelding 2026-01-29 om 20 12 49


```css
diff --git a/static/style/readme.css b/static/style/readme.css
index 70749ce4..343162c3 100644
--- a/static/style/readme.css
+++ b/static/style/readme.css
@@ -131,13 +131,22 @@
   .markdown-anchor {
     opacity: 0;
     position: absolute;
-    left: -1em;
-    bottom: 0.1em;
+    left: -1.2em;
+    bottom: 0.3em;
     color: var(--foreground-3);
-    width: 1em;
+    width: .9em;
+    height: .9em;
+    line-height: 1;
+    justify-content: center;
+    align-items: center;
+    display: flex;
+
     text-decoration: none;
-  }
-  .markdown-anchor:after {
-    content: "⎆";
+
+    &::after {
+      font-size: .66em;
+      content: "⌘";
+      display: block;
+    }
   }
 }
```

By the way, the anchors need a label (for accessibility) so perhaps add `aria-labelledby` pointing to the parent's id.

static/style/readme.css
braver: I do like the darker orange for the notes and IMO doesn't conflict with the links being in the "normal" orange (since they're just really different and unrelated things).

Image

static/style/readme.css
```diff
@@ -74,4 +74,70 @@
...
+      border-left-color: hsl(29.62deg 66.08% 57.86%);
+    }
+
+    :root[data-theme='dark'] & {
+      &.markdown-alert-note,
+      &.markdown-alert-tip,
+      &.markdown-alert-important,
+      &.markdown-alert-warning,
+      &.markdown-alert-caution {
+        border-left-color: hsl(32.77deg 70.44% 55.43%);
```
braver: Note that you can probably just nest `:root[data-theme='dark']` in the previous selector, instead of repeating all the alerts styles here. Or of course introduce a new color variable in colors.css and set up the dark variant there.

static/style/readme.css
braver: I also don't mind our `--blue-3` here:

Image

braver:
> in theory, different note types(?) could have different colors

Yeah, I don't see a real need there as long as we avoid obvious signal colors (red/green). It's clear enough to call attention to itself by just existing, I don't think need to make it any more complicated than that. I hardly ever see them in the wild anyway.

static/style/readme.css
kaste: Maybe we just stick to elegant and readable.   E.g.  
image
 
I picked the note color rather quickly, just to have a contrast to the black and white text.

static/style/readme.css
```diff
@@ -74,4 +74,70 @@
...
+    opacity: 0;
+    position: absolute;
+    left: -1em;
+    bottom: 0.1em;
+    color: var(--foreground-3);
+    width: 1em;
+    text-decoration: none;
+  }
+  .markdown-anchor:after {
+    content: "⎆";
```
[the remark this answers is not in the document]
kaste: image
image
image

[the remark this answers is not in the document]
braver: `#` definitely works for me :)

static/style/readme.css
kaste: > Links do need to have a color though,

Not really, it is just focus towards readability and a calm view.  I got that design from German newspapers for what it's worth.   Plenty of sites neither have a color nor underlines too.   ...  I just don't like the wording I guess.

The original design actually was blue -> https://github.com/packagecontrol/thecrawl/issues/296#issuecomment-3772117349 I don't think that's a reasonable option to be honest, I don't want to switch to blue and have the orange in the footer and in the metadata area above the readme.

static/style/readme.css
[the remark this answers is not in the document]
braver: Haha, also fine 👌🏻

braver:
Speaking of newspapers, here’s an example that uses color, but it’s balanced. The red goes with the blue and the blue doesn’t interrupt the reading of the text. We don’t have a color set that does that: the orange is too hot and the other colors are based off of it.

image

kaste:
I'll close this then and just merge the technical bit.

PR #333, opened 2026-02-04
Draw red dots for all releases in the current chart

kaste:


kaste:
429 from Cloudflare, and then we fall back and fetch the logs from our site.  (Because I'm so absolutely smart and did that fallback.)  These are of course only as fresh as the build.

PR #336, opened 2026-02-05
feat: add top navigation bar for better discoverability

### LINKED ISSUE #335: Make purpose of the site more obvious
Hi all,

I don’t use sublime often, and was used to the old (kind of defunct) package control page. 

### Problem: 
I stumbled across this page several times but kind of ignored it because information gathering was not obvious. I never scrolled to the very bottom to discover the “About”, “Submit your package” and so on links; the info is a bit hidden.

Only later I realized that this is the new site, and read all the content under the About link.

### Solution:
Add a simple navbar at the top, users can click and read the content of About , and get re-directed to Submit your package or the docs immediately.

Can be a navbar at the side too. But somewhere obvious would be great !

thanks

ilyaZar:
Fix #335

Just for illustration purposes. Some people may prefer such navbar information on the side. Branch for illustration is https://ilyazar.github.io/thecrawl/

For the moment at the top as a short minimal fix , so people see whats going on

# Light:
image

# Dark
image

cheers

### DIFF
--- a/_includes/layout.njk
+++ b/_includes/layout.njk
@@ -4,6 +4,7 @@
 
   <body>
     {% include "header.njk" %}
+    {% include "site-nav.njk" %}
     {% include "templates.njk" %}
 
     <main>
--- a/_includes/site-nav.njk
+++ b/_includes/site-nav.njk
@@ -1,0 +1,15 @@
+{% import "util/macros.njk" as util %}
+
+<nav class="site-nav" aria-label="Site navigation">
+  <div class="site-nav-links">
+    <a href="/about">About</a>
+    <a href="https://docs.sublimetext.io/guide/package-control/usage.html">Using Package Control</a>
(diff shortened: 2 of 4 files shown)


kaste:
LOL, yeah man.

braver:
SublimeHQ has an excellent track record of working _with_ the community on core aspects of the project, like [packages](https://github.com/sublimehq/Packages), and the community docs are respected and listed on the main website. So I'm sure we'll come to a balance that works for everyone involved. 

However, it's currently just not very clear where each individual component of this project will land. They're probably mostly going to be concerned with the underlying infrastructure and the package itself, not so much with the website. But we'll see. 

With that, how we solve #335 will need a different approach depending on the outcomes of that.

Moving links to top of the page is not the answer IMO, they're fine in the footer. We used to have "about" in the top bar, we could bring that back. But right now the purpose of the site being unclear is more caused by the project being in a transition period than any particular design decision.

PR #351, opened 2026-03-31
Persistent registry; lifecycle management

kaste:
Let's try this one.  

Created an orphaned branch the-registry, changed crawl.yml to read the registry from there, and to push it (if it changed) to that branch back.  

generate_registry learned lifecycle management.  New packages are tagged with "first_seen"; removed packages are re-added as tombstones.
 
A new tool `generate_seed.py` extracts "first_seen"/"removed" data from workspace.json files or registries.  

Old data has been scraped from packagecontrol.io; so we have more tombstones in the database than before.  Unlikely we have all; don't know what the actual policy was for pc.io.


braver:
> don't know what the actual policy was for pc.io.

Not sure what it looks like in the data, but on the site removals are simply not handled (ie. nothing gets removed or even marked as such). _Maybe_ sometimes Will did something manually sometimes but not in recent years. I removed this one ages ago for instance: https://packagecontrol.io/packages/Theme%20-%20Sea%20Lion.

kaste:
Oh, they just look like pages that don't get downloads.  How would I scrape them?  It is basically just a page with an outdated LAST SEEN tag, so you go just through all the pages and look for that.  Certainly possible.

kaste:
Actually, this is typically to be installed on the package_control_channel repo; but we don't have enough permissions to run it over there and it'd be just a hassle.

braver:
Maybe one day though! 🤞🏻

PR #352, opened 2026-04-08
fix #82 make hand-picked labels stand out

### LINKED ISSUE #82: Some labels should stand out...

braver:
The discussion in #82 ended up going towards "just" using bold (and a slight color contrast adjustment) to highlight the hand-picked labels. We've gone a few steps way from colorful lately and I think that's a good thing after all. Also, the language icons/logos already add color. Bold does what it needs to do while also being completely inoffensive. 

I considered adding LSP to that list, but it feels a bit random because that's not part of the selected labels we put under the search field. Besides, all LSP packages already have LSP- in their name, so LSP is written all over them to the point that highlighting the label feels unnecessary. 
I also considered adding AI to that list, but IMO it's not that kind of label right now. We're in the middle of the hype cycle, perhaps we'll end up with more of a SublimeLinter/LSP kind of situation around interacting with LLMS and agents. Right now it can mean anything.

So, tiny little PR:

Scherm­afbeelding 2026-04-08 om 10 55 54

Scherm­afbeelding 2026-04-08 om 10 58 12

Scherm­afbeelding 2026-04-08 om 11 06 32

### DIFF
--- a/eleventy.config.mjs
+++ b/eleventy.config.mjs
@@ -260,13 +260,18 @@ function basePackage(pkg, stat) {
     return minB - minA // Higher min build first
   })
 
-  const labels = pkg.labels?.slice() ?? []
-  if (!supportsModernSublime) labels.push('ST2')
-  if (supportsModernSublime && doesNotSupportNewestSublime) labels.push('ST3')
+  let labels = pkg.labels?.slice() ?? []
+  labels = labels.sort((a) => {
+    if (['language syntax', 'snippets', 'linting', 'auto-complete', 'color scheme', 'theme'].includes(a)) {
+      return -1
+    }
+  })
+  if (!supportsModernSublime) labels.unshift('ST2')
+  if (supportsModernSublime && doesNotSupportNewestSublime) labels.unshift('ST3')
   if (pkg.removed) {
-    labels.push('RIP')
+    labels.unshift('RIP')
   } else if (pkg.archived_at) {
-    labels.push('MIA')
+    labels.unshift('MIA')
   }
 
   const platforms = util.computePlatformLabelsForSearch(rawReleases).sort()
--- a/static/style/buttons.css
+++ b/static/style/buttons.css
@@ -87,7 +87,20 @@
   &.label-more:hover {
     color: var(--orange-2);
   }
+
+  &.label {
+    &[href*='language%20syntax'],
+    &[href*='snippets'],
+    &[href*='linting'],
+    &[href*='auto-complete'],
+    &[href*='color%20scheme'],
+    &[href*='theme'] {
+      color: var(--foreground-1);
+      font-weight: bold;
+    }
+  }
 }
+
 
 span.button {
   cursor: revert;
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -108,7 +108,7 @@ section[name="labels"] .grid {
     top: 18px;
     right: 1rem;
     gap: 0;
-    font-size: 13px;
+    font-size: 11px;
     background: inherit;
     color: var(--foreground-3);
     svg {
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -108,7 +108,7 @@ section[name="labels"] .grid {
     top: 18px;
     right: 1rem;
     gap: 0;
-    font-size: 12px;
+    font-size: 13px;
     background: inherit;
     color: var(--foreground-3);
     svg {
@@ -118,7 +118,7 @@ section[name="labels"] .grid {
   }
 
   .labels {
-    font-size: 12px;
+    font-size: 13px;
     justify-content: flex-end;
     padding: 0;
     margin: 0;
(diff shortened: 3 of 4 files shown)


braver:
For your consideration, a second commit that moves the important labels to the front. That way they stand out more, and makes lists of similar packages easier to scan. It also just looks a bit cleaner. The handpicked labels are sorted to the front of the list, and our "synthetic" labels (ST2/3, MIA, RIP) are inserted at the front of the list.

Before:

Scherm­afbeelding 2026-04-08 om 11 21 38

After:

Scherm­afbeelding 2026-04-08 om 11 21 47

Also:

Scherm­afbeelding 2026-04-08 om 11 24 59

eleventy.config.mjs
```diff
@@ -260,13 +260,18 @@ function basePackage(pkg, stat) {
     return minB - minA // Higher min build first
   })
 
-  const labels = pkg.labels?.slice() ?? []
-  if (!supportsModernSublime) labels.push('ST2')
-  if (supportsModernSublime && doesNotSupportNewestSublime) labels.push('ST3')
+  let labels = pkg.labels?.slice() ?? []
+  labels = labels.sort((a) => {
+    if (['language syntax', 'snippets', 'linting', 'auto-complete', 'color scheme', 'theme'].includes(a)) {
```
michaelblyons, line 265: Do you want to change the sort to have these labels in a predictable order rather than just reversed-but-at-the-front?

braver: Yeah I was thinking about that and couldn't decide really... I think I'd want to keep labels in the order they were applied to the package by the author, but these 6 in front of all the others.

kaste:
That's interesting.  Bold/Italic is usually used to emphasize something in a sea of content, but then you move it at the front and you have something like a gradient.   E.g. first the synthetic red labels, then bold then normal.  

Since I always wondered about the sorting here, it falls naturally: sort by "popularity" of the label.  They fade out to the right, if not visually -- In real typography you could make the font actually thinner and thicker gradually -- then semantically.  

Doesn't have to be in this PR as while we have the data (label => usage count) somewhere to render the Labels page I'm not sure exactly where.  

Is the implementation hacky or clever?   Well, are these hand-picked or not really the six most used labels?  Very likely I would have defined a `.label.emphasized` CSS rule and assigned that to the most popular labels.  

LGTM, good idea.  However, on low DPI screen the (bold) font is a bit too packed/cramped/noisy:

image
image

and 13px would be easier to read:

image

braver:
> are these hand-picked or not really the six most used labels

Well, they're also the most used. But that there are 6 is almost random. And the selection is hardcoded and not based on the actual usage stats. 

The bold font does gets a bit chunky, let's see if I can tweak that.

braver:
Yeah, I don't hate it at 13 at all. Even the stats are more readable and less overly tiny. Let's go with 13:

Scherm­afbeelding 2026-04-09 om 20 20 14


By the way, as to this:
> Very likely I would have defined a .label.emphasized CSS rule and assigned that to the most popular labels.

That would perhaps have been better, but it would have to be done in both eleventy and JS and just feels "hard". Labels have a very predictable href, so I like to go with "clever" over "proper" sometimes 😄

kaste:
On low-DPI the right top corner is actually a drastic change:

image
image

<br>
image
image

The proportion to the icons change drastically; also the vertical whitespace between the two lines.  That's risky.

kaste:
Can you actually simulate low-DPI, @braver?  I learned that while doing the SVG charts (install and status chart); the rounding glitches are huge, e.g. you draw vertical lines for the grid and suddenly the lines have different widths or even have a slight angle.

braver:
> Can you actually simulate low-DPI

I don’t know that I can. All my devices and everything at the office are at least @2x so stuff just gets blurry in weird ways.

braver:
On those screenshots of yours though, the line height change and alignment shift is unfortunate but the the text looks a lot sharper at 13?

kaste:
image
<br>
image

I just took these ☝️ and honestly they all look blurred when pasting it here and better on the site.  The first is with 11px for the top right stats which is probably even the best choice on my setup.  With low-DPI (and on my setup then) the font does not grow proportionally at all and on the second pic (12px) the 3.3K looks compressed, I just never noticed it before.  This is basically our font, other fonts (e.g. the one here on GitHub) scale better.

FWIW.

braver:
It really different from font to font and from Mac to Windows. Sometimes you make something 1px larger and it just becomes taller while staying exactly the same width 🙄 I picked Lato mostly because that's also what sublimetext.com uses (and IMO it looks very "normal" and "good" most of the time). 

The stats look :chefs-kiss: at 11px, let's do that 👍🏻

kaste:
Just go with it.

PR #360, opened 2026-04-21
Stabilize featured label ordering

kaste:
Keep manually picked featured labels at the front while preserving an explicit order among them.

Move the sorting logic into util.sortFeaturedLabelsFirst() and add unit tests that verify:
- featured labels are ordered by the provided featured list
- non-featured labels keep their original relative order

### DIFF
--- a/eleventy.config.mjs
+++ b/eleventy.config.mjs
@@ -12,6 +12,15 @@ const supportedRepackagerHosts = [
   'https://codelab.org/',
   'https://gitlab.com/',
 ]
+const FEATURED_LABELS = [
+  'language syntax',
+  'snippets',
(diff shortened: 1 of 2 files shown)


kaste:
Before:

image

After:

image

kaste:
WDYT @braver @michaelblyons? Polls are open.

michaelblyons:
Love it. It's nice on the homepage but *extra* helpful in a search.

braver:
Awesome. 

At some point _every_ little detail of this site will have been thought out specifically and precisely :)

kaste:
Ok.  I'll take that as a yes.

PR #361, opened 2026-04-21
Add dynamic featured label shortcuts (Additive)

kaste:
Build dynamic featured labels under search using active label
filters and co-occurring label usage.

When no label filter is active, keep curated defaults. When labels
are active, show all active labels first, then suggest up to the
remaining slots (max 6 total suggested+active cap logic).
Exclude synthetic status labels.

Sister draft at #362

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -1,6 +1,25 @@
 import { List } from './module/list.js'
 import MiniSearch from 'https://cdn.jsdelivr.net/npm/minisearch@7.1.2/+esm'
 import { createMinisearch } from './module/minisearch.js'
+import {
+  appendFilterToken,
+  buildFeaturedLabels,
+  buildLabelRecords,
+  hasFilterValue,
+  parseSingleFilterQuery,
+  removeFilterValue,
+} from './module/search-query.js'
+
+const CURATED_FEATURED_LABELS = [
+  'language syntax',
+  'snippets',
+  'linting',
+  'auto-complete',
+  'color scheme',
+  'theme',
+]
+const DYNAMIC_LABEL_EXCLUSIONS = ['ST2', 'ST3', 'MIA', 'RIP', 'FAILING']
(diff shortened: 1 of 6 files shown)


kaste:
That just came to my mind, example


https://github.com/user-attachments/assets/61929f5c-f386-4c45-9ea4-d56cb6c1ed80

This is fun to click.  This is the "add" variant.  I also have a less fancy replace variant.  

@michaelblyons @braver

braver:
At first glance I like this one better than the other one. It invites you to click more labels to dig deeper, which makes more sense to me than "try this other label". But, having said that, in cases where there are multiple related or similar labels "try something else" might actually be a more common use case. In any case I really like the idea of using those 6 highlighted labels to suggest something dynamically. You could also do that based on search term perhaps, showing the 6 most common labels within the result set (like, oh you typed "lsp" but did you know there's a label like that you can use to filter?).

kaste:
The suggested labels are now based on the search query.  Looks freaking promising tbh.  

I actually exclude the label if the term is an exact match. E.g. you search "git" or "lsp" or "python", then every time the top label *was* the search term which is not a good suggestion.  So I exclude exact matches and this behaves 👩‍🍳.

braver:
> Looks freaking promising tbh.

💯 

I like this. Also better than the replacement thing (the other PR) because you're inviting to refine, to dig deeper, and that's what this does consistently. Although results don't become better once you add more than one label, that's typically too specific. But it feels like this makes the most sense to me.

kaste:
Let's play with this one.

michaelblyons:
Belatedly: This is cool! I don't use the tags that often from the homepage, but...

- It is fun to see what kind of results you are going to see when you type a search.

- It's also a little interesting to click one of the (more prolific) author names and you can see what kind of packages they maintain... assuming they have tagged their packages.

PR #362, opened 2026-04-21
Add dynamic featured label shortcuts - 2 (Replacing)

kaste:
Same as #361 but clicking labels isn't additive to the search query but always replaces the query.

Less confusing; the main feature here is that the featured labels below the search input now dynamically change to the most used labels that co-occur with the selected label.  

Example

https://github.com/user-attachments/assets/baf0415b-4b0a-4a22-ab89-bb0cea9769d2

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -206,9 +206,11 @@ document.addEventListener('click', (event) => {
     const oldQuery = input.value
     const clicked = parseSingleFilterQuery(newQuery)
 
-    if (clicked?.type === 'author') {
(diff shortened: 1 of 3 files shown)


kaste:
This is less confusing maybe; but also a bit less fun to just click around and "find" things, go down a rabbit hole, "narrow down a search" because the search list is always long.  There is no narrowing by just clicking.  You need to type.  But graphically it is great.  

@braver @michaelblyons again.

kaste:
We went with #361.

PR #363, opened 2026-04-22
Add source link with live repository URL

### LINKED ISSUE #355: Expose link to source `repository.json` on package page
Can you add a (maybe subtle) link to the JSON file where a package comes from? Sometimes I want to fix their metadata, but if it's not directly in PCC I don't know where to look.

kaste:
Fixes #355

Expose a new Source button in the Links section on package pages
using the log icon.

Convert raw source URLs to browsable/editable file URLs, including
GitHub raw URLs with refs/heads paths.

For the package_control_channel source URL, resolve the actual include
file by package name and rewrite the legacy wbond path to the canonical
sublimehq repository path.


michaelblyons:
Neat! I would have cheated the line indexing with [Text Fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments), but you did all the work for real line indexing already.

Is there a place upstream where `wbond` is used that can be changed to `sublimehq` and skip the transforms you have?

kaste:
I think it is only one transform here, and yes thecrawl itself only references `wbond`.  Likely this is the first use of `sublimehq`.  This is also politics actually.  I'm not sure if I should switch to sublimehq just now.

PR #364, opened 2026-04-24
Refine install chart release version label

kaste:
Adjust weekly release callout labels to keep short cases readable while still allowing access to full version lists.

For one version, show only that tag.
For two versions, show "v2 / v1".
For three or more, show a collapsed "v2 / v1 / ..." label by default. 

Expand to the full list when hovering the red release dot.


kaste:
E.g.

image

On hover:

image

PR #365, opened 2026-04-24
Prevent avg lines from stealing chart hover

kaste:
The install chart's horizontal average lines were receiving pointer hover, which could interrupt week/release hover state transitions.

Just set avg-line pointer-events to none.


kaste:
Flicker hunter.

PR #366, opened 2026-04-27
Ignore unknown active labels

kaste:
Only labels present in the dataset can be rendered as active featured labels.  This prevents incomplete label filters from creating shortcut buttons for labels that do not exist.

E.g. the user types `label:"powersh` (for "powershell").  Before this change we created a synthetic label "powersh".


kaste:
This prevents intermediate, fake labels, e.g.

image

kaste:
I thought I had fixed that yesterday but no I didn't.

PR #367, opened 2026-04-30
Fix newest version hiccup in the status chart

kaste:
We basically computed what `newestTagBeforeDayWindow` now does on build
time.  Naturally, this gets out-of-sync if the live chart advances
before a next rebuild.

The fix is to compute it "live", client-side.


kaste:
That was the bug:

image

Issue #56, opened 2025-07-30
Simplify platform tags if able


michaelblyons:
Can we collapse package tags when they fully cover a simpler combination? Like the "tornado" library has

- `osx-arm64`
- `osx-x64`
- `linux-arm64`
- `linux-x32`
- `linux-x64`
- `windows-x32`
- `windows-x64`

Which could be flattened, IIUC, to

- `osx`
- `linux`
- `windows-x32`
- `windows-x64`

And if ST doesn't run on `windows-arm`, it could just display no tags at all, yes?

Related to #12

michaelblyons:
Another possibility to shrink the tag duplication (and apologies if this is too much) is to fold the architecture into a subtag, where you can click the main platform of the tag or the arch if not fully spanned by all arch subtags:

- `linux-arm64` -->
    - `linux (arm64)`
- `linux-arm64`, `linux-x64` -->
    - `linux (arm64) (x64)` or
    - `linux (arm64|x64)`

each with `linux`, `arm64`, and `x64` its own `<a>`. Punctuation not meant to be actual characters. I'm imagining the background darkened inside the pill(s).

FichteFoll:
As I suggested in https://github.com/packagecontrol/thecrawl/issues/12#issuecomment-3029608479, this problem would also become more wieldy by showing labels of the platform and arch combinations that are *not* supported. I.e., show "!linux" if no linux arch is supported and only "!linux (arm64)" if support for x86 and x86_64 is there (notated as "x64").

michaelblyons:
> As I suggested in [#12 (comment)](https://github.com/packagecontrol/thecrawl/issues/12#issuecomment-3029608479), this problem would also become more wieldy by showing labels of the platform and arch combinations that are _not_ supported. I.e., show "!linux" if no linux arch is supported and only "!linux (arm64)" if support for x86 and x86_64 is there (notated as "x64").
> 
> Note: The exclamation mark is just for visualization here. How they are actually displayed is up for debate.

I like the `!linux` (however it is displayed), as I agree that's more relevant for someone actually looking for whether a package supports their OS/arch.

I do worry a little bit that we're going to find some weird edge cases where the tags kind of balloon unexpectedly. Let's take the Codecs26 and Codecs33 from your screenshot in #12:

- Codecs33: `⛔ windows` `⛔ linux (arm64)`
- Codecs26: `⛔ windows` `⛔ macos` `⛔ linux (arm64)`

[...]

PR #371, opened 2026-05-06
wrap platform statement if needed

braver:
Before:

Scherm­afbeelding 2026-05-06 om 20 42 05

After:

Scherm­afbeelding 2026-05-06 om 20 39 42

As reported by @michaelblyons in https://github.com/packagecontrol/thecrawl/issues/56#issuecomment-4390844413

### DIFF
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -135,6 +135,7 @@ section[name="labels"] .grid {
       padding: 0;
       cursor: default;
       white-space: revert;
+      word-break: keep-all;
       &:hover {
         background: transparent;
       }
@@ -144,6 +145,14 @@ section[name="labels"] .grid {
   .list & .labels {
     justify-content: flex-start;
     margin-top: 21px;  /* usually collapses with margin-bottom from .description */
+  }
+
+  .list & .platforms {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    display: inline-block;
+    max-width: max(100px, 30vw);
   }
 
   .list & .stats {
--- a/static/style/grid.css
+++ b/static/style/grid.css
@@ -134,6 +134,7 @@ section[name="labels"] .grid {
       margin: 0;
       padding: 0;
       cursor: default;
+      white-space: revert;
       &:hover {
         background: transparent;
       }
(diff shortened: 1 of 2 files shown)


kaste:
Doubt it


image


E.g. join by ` / `.

image
image

michaelblyons:
> E.g. join by `/`.
>  image

Yeah, the space joining is better.

kaste:
If you don't want to have more whitespace we could also join by a ZWSP (`U+200B`) and see if the browser prefers splitting on these space then.  

For the search page, you just need to guide me to a solution if you can't make the change.

braver:
This (screenshot below) is more of a problem than the splitting. Joining with ` / ` works for my eye anyway, but I'm partial to commas: they're more text-y and less graphic. We don't have to guess which works better across browsers, CSS should give us enough control.

But platform "statements" (ie. more than just one platform, or one not-platform) will never fit in that top-right corner.

Scherm­afbeelding 2026-05-11 om 21 24 24

<br>
If we're lucky and the package and author are short, sure, but not reliably.

Scherm­afbeelding 2026-05-11 om 21 27 20

<br>
Maybe just hide it at that point (and reveal in a tooltip)?

Scherm­afbeelding 2026-05-11 om 21 30 16

<br>
On detail pages this "fine" for me. It's not overly pretty but it doesn't hurt anyone for the handful of packages where the statement wraps.

Scherm­afbeelding 2026-05-11 om 21 35 58

<br>
I like this actually, with commas:
Scherm­afbeelding 2026-05-11 om 21 40 42

---

Anyway .... the search results list is still an issue after all that. Clipping is for now the best I can come up with. Perhaps the alternative would be to switch to the smaller cards that we use in grids on the homepage, on smaller screens. But LSP-file-watcher-rust becomes a problem on already relatively wide screens. Anyway, this isn't too bad. The ellipsis indicates there is more, and the full information is available on the detail page.

Large screen:
Scherm­afbeelding 2026-05-11 om 21 54 24

<br>
Small screen:
Scherm­afbeelding 2026-05-11 om 21 53 56

kaste:
Fixed in #372 

It is not super difficult to make it work in the search results too.  I try a refactor next.

PR #372, opened 2026-05-12
Improve platform statement wrapping

kaste:
Allow platform statements in card labels to wrap naturally and align
stats text on the package details page by baseline.

To make that work for the search results page, we really need a different structure.  (Short: Make the top of the card a two column layout.)

This need a refactoring PR later as we have "one" card that actually doesn't share that much, a compact one the landing page and a search result card/entry/item.  Also: platform statements are neither labels nor buttons.  These are non-interactive metadata text.  In the sense `white-space: revert` hacks around already.  

image

image

image

These all look better now aligned to the corner.

### DIFF
--- a/_includes/packages/macros.njk
+++ b/_includes/packages/macros.njk
@@ -55,12 +55,7 @@
 {% endmacro %}
 
 {% macro labels(pkg, with_icons=false, platform_statement='') %}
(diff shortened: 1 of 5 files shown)


braver:
Nice!

PR #375, opened 2026-05-19
Redo library listing cards

### LINKED ISSUE #373: lsp_utils library doesn't show version
https://packages.sublimetext.io/libraries?q=lsp_utils shows "package or version not found" in the `pypi` badge. Why is that? Is it because releases use the `v` prefix for tags?

kaste:
Render libraries from the workspace data so the listing uses current
versions, platform metadata, and descriptions.

Align library cards with search result cards, remove legacy shield and
platform tags, and show version and platform details in the card meta
area.

Fixes #373

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,6 @@ _site/*
 .package_control_channel/
 channel.json
 logs.json
-libraries.json
 registry.json
 stats.json
 workspace.json
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,6 @@ build:
 update-data:
 	# workspace from https://github.com/packagecontrol/thecrawl
 	curl -o workspace.json -L "https://github.com/packagecontrol/thecrawl/releases/download/crawler-status/workspace.json"
-	# libraries repository from https://github.com/packagecontrol/channel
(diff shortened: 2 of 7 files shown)


kaste:
That's a clean, innocent, minimal render

image

The important part is that we use *our* workspace data for the render.

braver:
Very nice!

kaste:
I have actually more information/details to show.  Matter of design.

image

image

kaste:
That's an adaption of the CLI tool output

```
$ uv run -m scripts.crawl_libraries --name arrow
arrow release matrix; -v to see the raw JSON output
Source: pypi:cache-304
Latest version: 1.4.0

             py33  py38  py313  py314
windows-x64  A     B     B      B
windows-x32  A     B     B      B
osx-x64      A     B     B      B
osx-arm64    A     B     B      B
linux-x64    A     B     B      B
linux-arm64  A     B     B      B

A = 0.14.2
B = 1.4.0
```

michaelblyons:
Maybe each library can have a page like each package does? Instead of clicking directly to the GitHub or PyPI?

This makes more questions. (Is there a search at the top? Does it search libraries instead, or the normal search? Etc.) But there would be space for the OS/Arch grid. And also space for a ReadMe (assuming PyPI exposes it).

PR #385, opened 2026-06-03
Add status chart update mode

kaste:
Add a text control for switching the status chart into update mode.

The mode recolors normal runs to grey and highlights runs with found
updates in green.


kaste:
This basically renders

image

image

FichteFoll:
What do the edges/lines mean?

kaste:
Point to the actual package updates; basically update happened to update discovered or vice versa.

PR #390, opened 2026-06-08
Tweak status chart prompt

kaste:
Use shorter wording for the status chart mode toggle.

### DIFF
--- a/status.njk
+++ b/status.njk
@@ -8,7 +8,7 @@ page_type: status
 <div class="status-chart-shell">
   <div class="status-chart-toolbar">
     <button type="button" class="status-chart-mode-button" data-status-mode-toggle aria-pressed="false">
-      Which runs actually did discover new updates?
+      Which runs actually found updates?
     </button>
   </div>
   <div class="status-chart" data-status-chart aria-hidden="true">


kaste:
🤷 @michaelblyons What needs to be done...

kaste:
Actually, I would prefer the/a longer version on wide screen.  It was just too wide on mobile.

michaelblyons:
The new text is better English as well as being shorter. If you want to emphasize some part of the message I can try to workshop the text.

Some possibilities:

- Which runs found updates?
- Which runs discovered updates?
- Which runs actually found updates?
- When were updates discovered?
- Mark runs with updates
- Mark runs with package updates
- ☑️ Runs with package updates

The last one might need a different word for "runs." It is more often used as a verb, and here it's a noun in a sentence fragment.

kaste:
Which runs actually did discover new updates?
    

      Which runs actually found updates?
    

> The new text is better English

I was aware of that when writing.  But also forgot before committing, or didn't
want to spend too many time with it anymore.

The initial momentum on that day was very graphical.  And after drawing icons
etc I noticed that I need something lengthy and wide.  But if I don't have
icons/colors, I only have length to guide the eye and to slow down
processing/reading I added that bit of clumsiness.  Compared to
"Which runs found updates?" which you just read without any processing, I basically
added slow-down words and rhythm.  I intentionally add friction to grab attention.

We need that friction so that the sentence does not carry "Which runs found
updates:" as if describing the currently visible chart.  We first see a noisy
chart, and then the button is a filter to emphasize some dots to answer a very
specific (Wald vor lauter Bäumen) question.  (Plus it transforms the chart
into an even more aethetical, graphical piece when it morphs into a modular
synthesizer picture, with the lines as cables.)

Too much friction is probably

      Which runs actually resulted in newly discovered updates?

Here area lot of dots.  Each dot represents, stands for a crawler run which is
a polling service.  Click here to highlight the dots where the crawler did not
run unnecessary, wasting CO2, but actually did find changes that resulted in a
new build of the site and the channel.

Also note the diagonal

image

which I need and is lost. 

image

(However, the short form is important for mobile!)

PR #393, opened 2026-06-09
Rename sort by "Magic" to "Relevance"

BenjaminSchaaf:
Magic is meaningless here, and the sorting is done by relevance.

### DIFF
--- a/util/macros.njk
+++ b/util/macros.njk
@@ -61,7 +61,7 @@
       <div class="button">
         <label for="sort-field">Sort by</label>
         <select name="sort" id="sort-field">
-          <option value="relevance" selected>Magic</option>
+          <option value="relevance" selected>Best match</option>
           <optgroup label="Popularity">
             <option value="installed">Installs</option>
             <option value="stars">Stars</option>


kaste:
I agree the default sort label can be clearer, but "Magic" was
intentional rather than meaningless. It dates back to the initial
default ranking work and refers to the site-specific ranking heuristic.

"Relevance" is the standard term, but this ranking is a hybrid of search
match score and package metadata. Maybe "Best match" is a better label:
less surprising than "Magic", but still broad enough for the actual
ranking behavior.

BenjaminSchaaf:
I didn't mean to insinuate that the word choice was unintentional, merely that magic is meaningless to a layperson. "Best match" seems good.

Probably obvious, but I'm trying to upstream my changes as we deploy the website officially :)

FichteFoll:
> "Relevance" is the standard term, but this ranking is a hybrid of search
> match score and package metadata.

To me, this sounds like what I would expect "Relevance" to mean.

BenjaminSchaaf:
I've pushed "Best match"

kaste:
I just cherry-pick.  I changed my mind and took your old "Relevance" -- which I also always used internally and in my PRs.  

I do get the initial intention of "Magic" as the relevance algorithms are in fact "magic" to the end-user, -- that's likely also the spirit of the ranking easter-egg we still ship: "Hej, here is the ranking transparent." -- and also Relevance is becoming a weasel word as all top players defaults now present irrelevant results as top matches, so the saying at least.  

But that pun is maybe too implicit and does not sparkle.  

"Best match" just tries to avoid the standard term, and has a space in it which is disliked visually.  It isn't punchy, just different to be different.  So I cherry-pick the previous commit with "Relevance".

PR #394, opened 2026-06-09
Optimize search index by removing redundant fields

BenjaminSchaaf:
The fields are ordered by how frequently they're empty, and any fields that are empty are removed back to front. This reduces the data stored in the search index from 987KB to 800KB (360KB to 325KB gzipped)

### DIFF
--- a/_includes/head.njk
+++ b/_includes/head.njk
@@ -14,6 +14,7 @@
   {% inline_js 'static/_prefetch_links.js' %}
   <script>
     window.STATIC_BASE = '/{{ "static/" | bust }}';
+    window.IS_PROD = {{ site.isProd }};
   </script>
 
   <link rel="stylesheet" href="{{ '/static/styles.css' | bust}}">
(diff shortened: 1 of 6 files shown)


kaste:
Thanks, there are a couple of good ideas here, especially deriving the
package href from `name` instead of carrying `permalink`.

I’m going to keep that part, but not the broader tuple trimming in this
form. `magic` is intentionally present for the debug comment/easter egg
-- I just touched this code yesterday -- and empty platforms are not 
necessarily equivalent to `any`.

If we keep the easter-egg the gzip savings drop from ~31.8 KiB to
about ~1.5 KiB when implemented carefully.

The client deserializer also needs another pass:
`window.isProd` should be `window.IS_PROD`, the condition does not match the
serializer, and `row.shift()` reads the package name rather than the magic
breakdown. This silently disables or corrupts the magic comment rather than
safely optimizing it.

Also, in ``permalink: `/packages/${encodeURIComponent(name)}``, `name`
is unbound.  

I keep the permalink idea and look into the `sort.js` cleanup.  I agree that if 
we normalize numeric fields while writing the index, the sort code can
stop reparsing them.  #395

PR #396, opened 2026-06-11
Update channel URLs to their new location

BenjaminSchaaf:
The repository has moved to sublimehq


kaste:
This one likely needs a db migration.  Need to think about it.

*EDIT*: Pretty sure we need a migration -- although it would (likely) self-heal over time -- because the channel would only refer the new channel URL.

PR #398, opened 2026-06-11
Optimise and improve label icons

BenjaminSchaaf:
This makes the following changes to improve/optimise labels:

Label icons with one usage are not included, saving 85KB compressed. These labels are already excluded from the label list, so there's not much point having the icon.

Excluded some label icons that are just the label in a fancy color. Having the label for C++ read as "C++ C++" just doesn't look good.

Deduplicated labels based on capitalization. "C" and "c" should not be different labels.


kaste:
I cherry-pick "Dedup labels" straight.  Note however, that its implementation needs an improvement as you just implemented "the first casing wins".  And we're lucky with "c" and "c++", but not so with e.g. "sublimelinter" and "SublimeLinter".  I think a "most *used* spelling/form wins" is a good heuristic and still dynamic.

> Label icons with one usage are not included, saving 85KB compressed. These labels are already excluded from the label list, so there's not much point having the icon.

That's the wrong claim, we have them everywhere:

image
image
image

I do agree that the three c-family icons look redundant. At that time, I actually also thought about omitting the text part and to keep just the icon. Another way is of course to just get better icons for these three.  

I do like, or rather would like, because we're not there yet, if all languages had their own icons.  E.g. 

image
vs
image
  
I miss them.  If I focus on their specific look, I hate them.

BenjaminSchaaf:
> That's the wrong claim, we have them everywhere

I meant the overall list of labels.

It's only a single package that has each of those labels, I don't think it's worth downloading 85KB for individual packages.

kaste:
> I meant the overall list of labels.

The first screenshot I posted is from the labels page.   

>  I don't think it's worth downloading 85KB for individual packages

The *current* download size is 85kB total; your 85kB refer the *un*compressed diff.  The savings are roughly 37kib.

BenjaminSchaaf:
> The first screenshot I posted is from the labels page.

The screenshot is from the search, not from the overall list of labels.

> The current download size is 85kB total; your 85kB refer the uncompressed diff. The savings are roughly 37kib.

Hmm, could've sworn I checked the gzipped size, but I clearly didn't. Still don't think that's a price worth paying for the label icons; we're considering removing the icons all-together for the official site.

kaste:
> overall list of labels

initial load, yes sure.  But the "Omitting labels ..." changes and is clickable as well. 

> we're considering removing the icons all-together 

I don't get the size thing at all (if it's not huge).  Either you have icons or not, and then: if you load them eager you always "overload", and if you load them lazy it looks shit or is a bit non-trivial to implement.  

I can relate to "none at all" much more.  *Personally*, I was never a fan, I just implemented it.  So -- so sorry I forgot the name of the original Sublime Text author -- if he chimes in and leaves a vote, we had 3 against 2.  

-- 

For the C family, omitting the text and just showing the icon, looks like this

image

which is an improvement if we keep the icons.  

-- 

I didn't go through the icons actually, but maybe we could just drop all non-language icons; e.g. I saw "audio" but we're a programming community.

BenjaminSchaaf:
> initial load, yes sure. But the "Omitting labels ..." changes and is clickable as well.

Interesting, it does not look clickable haha

> For the C family, omitting the text and just showing the icon, looks like this
> which is an improvement if we keep the icons.

I disagree, the C++ icon is hard to read as text but still looks awful next to C++ as text. Having only the icons also makes it invisible to native search.

> So -- so sorry I forgot the name of the original Sublime Text author -- if he chimes in and leaves a vote, we had 3 against 2.

Fwiw it was Jon's idea to remove the icons entirely after I showed him the C/C++ ones.

braver:
> Excluded some label icons that are just the label in a fancy color. Having the label for C++ read as "C++ C++" just doesn't look good.

I'd have to agree with that one. Keeping the icon but dropping the label is not a good move IMO. You can't do that for 99.9% of the other labels anyway so it would be weirdly inconsistent. 

There is probably a better C++ logo out there that _would_ work. Like this one:
https://commons.wikimedia.org/wiki/File:ISO_C%2B%2B_Logo.svg

kaste:
This has been superseded by #449.  Read esp. https://github.com/packagecontrol/thecrawl/pull/449#issuecomment-5046009463 for the rationale; the first important distinction is that #449 implements a dynamic solution and reacts automatically when the dataset changes; second, it doesn't throw away low usage icons but loads them if ever needed lazily -- that's likely something we can agree on -- the actual sprite size is even lower than here, we ship only 90 icons in the first sprite which fulfills 93% of the whole dataset; actual split parameters are configurable.

PR #400, opened 2026-06-12
Update channel URLs to their new location

kaste:
These repositories have moved to sublimehq

Follow up of #396 with a small migration commit, that can be reverted if applied once.

Closes #396


kaste:
image
@BenjaminSchaaf
💪

PR #403, opened 2026-06-15
Remove orientationchange listener

BenjaminSchaaf:
Not only has this been dropped by everything except WebKit, the resize event already gets triggered when orientation changes.


kaste:
That's correct.

PR #405, opened 2026-06-15
Don't display search clear button when input field is empty

BenjaminSchaaf:
On small screens the clear button cuts off the placeholder text, so lets not have it there at all.

This only sets `display: none` when the placeholder is shown, otherwise the button is just hidden when the input field is not focused.


kaste:
Good finding.

PR #406, opened 2026-06-15
Fix package version title pluralization

BenjaminSchaaf:
It should still be titled "Versions" even if those versions are hidden under "more".


kaste:
Yes, these were under different sections some time ago.

PR #407, opened 2026-06-15
Add environment variable for disabling link prefetch

BenjaminSchaaf:
We don't want to use this for the official packagecontrol website for bandwidth reasons. It also doesn't work on firefox nor is it relevant for mobile.


kaste:
Drop the "feature"

kaste:
I cherry-pick this one; but really just revert the introducing commit (glad I found it and it is self-contained in just one commit).

PR #408, opened 2026-06-15
Swap download and release notes links

BenjaminSchaaf:
Getting a download when clicking a regular link is unexpected. Downloads should always be identifiable. The easy fix here is to swap the links around, so the version goes to the release notes and the icon is a download icon.

### DIFF
--- a/static/style/package.css
+++ b/static/style/package.css
@@ -45,7 +45,13 @@ html.page-package {
 }
 
 .download {
-  font-weight: bold;
+  color: var(--foreground-2);
+  vertical-align: -2px;
+  margin-left: 2px;
+
+  &:hover {
+    color: var(--orange-2);
+  }
 }
 
 .download-notes-sep {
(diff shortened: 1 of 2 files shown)


static/style/package.css
```diff
@@ -45,7 +45,13 @@ html.page-package {
 }
 
 .download {
-  font-weight: bold;
+  color: var(--foreground-2);
+  vertical-align: -2px;
```
kaste, line 49:
```suggestion
  vertical-align: -1px;
```

The notes icon just had a different shape.  I think -1 is enough.

kaste:
Yes, I actually thought we can move both targets to an icon -- because "notes" is leaving our site which is kinda surprising for just a different colored text.  

But this change is in the direction of it.  Just the alignment nit.

michaelblyons (approved):

kaste:
I just cherry pick this one to avoid the wait for a rebase.

Issue #423, opened 2026-06-17
Normalize labels


kaste:
The base for https://packages.sublimetext.io/packages/Modula-2%20Language%20Syntax is 

<img width="846" height="419" alt="Image" src="https://github.com/user-attachments/assets/f67d7f30-2aa1-4399-ac03-c6004aa1d13f" />

Note that `auto complete` is a spelling/format variant, so we map it to our common canonical form `auto-complete`.  

<img width="865" height="370" alt="Image" src="https://github.com/user-attachments/assets/9a71ad31-4b1b-4e04-a2e0-e362505c9c64" />

* This rewriting of the labels is easy enough, `cpp -> c++; innosetup -> inno setup` and so on.
* I would like to be transparent about this so people know, understand, can criticize us, or change their labels.

That is an UI problem.  I already have a map ready at render time which I transform to the clunky sentence in the picture.   

@michaelblyons 👀

kaste:
Classical other mocks

<img width="809" height="233" alt="Image" src="https://github.com/user-attachments/assets/73fd54cf-d5e5-4f58-a9ab-838a81184be7" />

<img width="865" height="370" alt="Image" src="https://github.com/user-attachments/assets/b7b6391c-a1f9-415c-aad3-50b8b3f5936d" />

<img width="458" height="346" alt="Image" src="https://github.com/user-attachments/assets/22118686-55d6-463a-a959-528c4c097833" />

michaelblyons:
That's awesome!

I like the last mockup the best. I'm not sure if it was intended to be for mobile only, but I think I like it for both.

On [the Labels page](https://packages.sublimetext.io/labels), will those be normalized, too? (and is it worthwhile to have an unfixed version? I might be the only person who cares about that, and I will care somewhat less if you have this map.)

Last very separate thought: Maybe add a warning on st_package_reviewer for using labels that are mapped to other labels.

kaste:
Right now I make that directly at ingest-data-level so the labels disappear everywhere, also on the labels page.  Now that I'm writing this, of course some hint/marker for labels that get normalized could be also cool.  Then we two design problems. :grin:

After patch:

<img width="835" height="187" alt="Image" src="https://github.com/user-attachments/assets/539b8fc9-2a9f-426c-989a-80ea6cb35f2b" />

Before:

<img width="828" height="183" alt="Image" src="https://github.com/user-attachments/assets/c0e0defe-7680-4197-838c-923e85989a6d" />

But there is also

[...]

PR #425, opened 2026-06-18
Add more label aliases

michaelblyons:
See #423

More possible candidates:
- `continuous integration` (which maybe should get `auto-test` instead of `testing` having it...?)
- `snippets` <- `snippet`
- ~~`c#` <- `c-sharp`~~
- Something with the `logs` and friends
- The various spellings of `brazil`
- `cheat sheet`
- `distraction-free`
- things that appear as `foo.js` and `foojs`
- `codestyle` / `code style`?
- `code generation`
- `convert` and friends
- `gfm` / `github *markdown`
- things that come up when you search for `duplicate`
- `ruby gems` / `rubygems`

### DIFF
--- a/label-aliases.json
+++ b/label-aliases.json
@@ -26,6 +26,12 @@
   "cheat sheet": [
     "cheatsheet"
   ],
+  "code generation": [
+    "code-generation"
+  ],
+  "coldfusion": [
+    "Cold Fusion"
+  ],
   "color": [
     "colour"
   ],
@@ -41,6 +47,9 @@
     "completion",
     "javascript completions"
   ],
+  "continuous integration": [
+    "continuous-integration"
+  ],
   "dark": [
     "dark colorscheme"
   ],
@@ -50,11 +59,23 @@
   "email": [
     "e-mail"
   ],
+  "file merge": [
+    "filemerge"
+  ],
+  "frontend": [
+    "front-end"
+  ],
+  "full screen": [
+    "fullscreen"
+  ],
   "inno setup": [
     "innosetup"
   ],
   "javascript": [
     "js"
+  ],
+  "key binding": [
+    "keybinding"
   ],
   "language syntax": [
     "langage syntax"
@@ -62,14 +83,26 @@
   "lorem": [
     "loremipsum"
   ],
+  "mivascript": [
+    "miva script"
+  ],
   "openstreetmap": [
     "osm"
   ],
   "parentheses": [
     "parenthesis"
   ],
+  "plain text": [
+    "plaintext"
+  ],
   "python": [
     "pyton"
+  ],
+  "rpgmaker": [
+    "rpg maker"
+  ],
+  "ruby gems": [
+    "rubygems"
   ],
   "snippets": [
     "snippet"
@@ -84,11 +117,17 @@
   "syntax highlighting": [
     "syntax highlight"
   ],
+  "template toolkit": [
+    "templatetoolkit"
+  ],
   "testing": [
     "auto-testing",
     "django-test",
     "python-test",
     "test",
     "unittest"
+  ],
+  "title bar": [
+    "titlebar"
   ]
 }
--- a/label-aliases.json
+++ b/label-aliases.json
@@ -3,9 +3,6 @@
   "addon": [
     "add-on"
   ],
-  "agent": [
-    "agents"
-  ],
   "amx mod x": [
     "amxmodx"
   ],
@@ -14,9 +11,6 @@
     "autocomplete",
     "autocompletion",
     "erlang autocompletion"
-  ],
-  "bookmarks": [
-    "bookmark"
   ],
   "c#": [
     "c-sharp",
@@ -29,7 +23,6 @@
     "cheatsheet"
   ],
   "color": [
-    "colors",
     "colour"
   ],
   "color scheme": [
@@ -73,6 +66,9 @@
   ],
   "python": [
     "pyton"
+  ],
+  "snippets": [
+    "snippet"
   ],
   "sorting": [
     "sort"
--- a/label-aliases.json
+++ b/label-aliases.json
@@ -3,17 +3,43 @@
   "auto-complete": [
     "auto complete",
     "autocomplete",
-    "autocompletion"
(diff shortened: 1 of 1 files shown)


kaste:
The failing was the trailing test;  I just fixed it but now see that you actually marked it as draft.

I removed "tdd" as that seems a very specific flow/technique to me.  _EDIT: That can be discussed and re-added of course. I misread the process and wanted to merge the safe bits directly.  So sorry for being hasty._

Yes on the others.  "logs" is to difficul[t to tell.  logs if probably logfile/logfiles, but logging and logger could be separate things.  Hard to tell.  Same is: debug and debugger and debugging.  We only know that something like "php debug" is wrong as that should be two tags/labels. 

Generally, I think that is a good direction, also to show that the webpage should be useful for maintainers too.  We can collect a translation map here; but then we can also use that in the maintenance process: use it in the reviewer, re-format the registry automatically if we have a good list.

label-aliases.json
```diff
@@ -3,17 +3,43 @@
...
   "color scheme": [
-    "colorscheme"
+    "colorscheme",
+    "colour scheme"
+  ],
+  "comparison": [
+    "compare",
+    "directory comparison",
+    "file comparison"
+  ],
```
kaste: That's a bit uncommon, as a tag "compare" looks more typical than "comparison" but I did not check against the actual packages that use that, compund "X comparison" are always odd but these more so

label-aliases.json
```diff
@@ -23,7 +49,37 @@
...
+    "auto-testing",
+    "django-test",
+    "python-test",
+    "test",
+    "unittest"
+  ],
+  "translate": [
+    "translation",
+    "translator"
+  ],
```
kaste: Here you have translate (☝️ compare) and not translation.  Both are really different things and I again did not look up the actual packages.

michaelblyons: I'd rather use `translation` than `translate`. I think noun labels are better than verb labels. Adjective labels are more debatable in my mind.

Here I picked the one that had 7 existing uses over my normal preference. I'd be happy to use `translation` instead.

kaste:
We need a minimal approach.  We cannot say I tend to use verbs, I like nouns, I prefer plural.   We need to search for the clear cases, like color/colour, auto complete with or without hyphen.  E.g. just a short script that normalizes the labels by eliminating " -" and see what labels we get  by that.  Also typos.  

Not sure if eml really is the same as email.  I think labels are meant to be tags, and these also follow, allow and often encourage free association. 

We need to think about this more.

michaelblyons:
> We need a minimal approach. We cannot say I tend to use verbs, I like nouns, I prefer plural. We need to search for the clear cases, like color/colour, auto complete with or without hyphen. E.g. just a short script that normalizes the labels by eliminating " -" and see what labels we get by that. Also typos.

Fair. The only thing I've been concerned about is labels with wildly divergent meanings. Things like a file extension that could be for one thing or something completely different.

I can either cut this PR down to the unambiguous ones, start a new one, or abandon.

> Not sure if eml really is the same as email. I think labels are meant to be tags, and these also follow, allow and often encourage free association.

All three packages with the `eml` tag are language syntaxes for email files. All three of them also have the `email` *and `e-mail` tags*. One or two other packages with the `email` tag also exist that I guess *send* an email somewhere.

> We need to think about this more.

Again, fair. I am happy either to push for making best practices for tag use or to let people do their free association.

IMHO it's at least as big a problem that a significant fraction of the packages have no tags at all.

kaste:
Look again.   IMO after thinking about it, focus should be searchability.  E.g. "cold fusion" vs "coldfusion", the latter allows "coldf" the former allows "fusion" to match, but "fusion" is not a nick name for cold fusion.  (And "Cold Fusion", we don't do right now anyway.)  E.g. "mivascript" vs "miva script" (the trademark is MivaScript I think).   "script" would match "miva script" but "mivas" would not match "miva script".  That's the tradeoff basically.  So I would say we lean towards "mivascript" which allows "mivas".  

In the same vein, removing django-test as that hurts searchability.  For these, we maybe want expansion "django-test" => "django", "test".  

The enumeration here is worthwhile IMO.  We could change the search/term behavior, then change "coldfusion" here to "cold fusion" etc.  and teach the tokenizer:

 ```js
   "miva script" -> ["miva", "script", "mivascript"]
   "rpg maker" -> ["rpg", "maker", "rpgmaker"]
   "front-end" -> ["front", "end", "frontend"]
   "cold fusion" -> ["cold", "fusion", "coldfusion"]
 ```

Then, we had here on the left hand side, the canoncical label we want to show to the people, and do the searchability transformation in/before minisearch.  

_EDIT:_  None of them *look* better though IMO.  "coldfusion" is okay here, not cold vs hot fusion, etc.

label-aliases.json
```diff
@@ -6,6 +6,10 @@
   "amx mod x": [
     "amxmodx"
   ],
+  "angular": [
+    "angular2",
+    "angular.js"
+  ],
```
michaelblyons, line 12: Angular basically split into two languages:

- One is Angular 2 and higher, and typically called Angular.
- The other one, Angular 1.x, is usually called AngularJS.

In the current set of labeled packages this distinction is moot, since the only thing labeled `angular.js` is RIP. But there are AngularJS packages in the registry... they're just not tagged.

michaelblyons: Maybe a similar situation is the PERL 5 -> Raku change, but the PERL people were willing to rename PERL 6 to Raku instead of keeping the name like Google did with Angular.

kaste: angularjs and angularts 😁

label-aliases.json
```diff
@@ -36,7 +36,7 @@
     "colorscheme",
     "colour scheme"
   ],
-  "competive programming": [
+  "competitive programming": [
```
michaelblyons, line 39: 🤦‍♂️ Sorry.

label-aliases.json
kaste: Good that we just changed the presentation layer and not the db.  

But "angular.js" is ugly  `=> angularjs`, and `angular2 => angular`.

PR #431, opened 2026-06-21
Add "Remarkable" section

kaste:
Add a home page section backed by magic score rankings with compact package and label links.

image


michaelblyons:
I like this concept. I hope to think of a better section title. "Remarkable" is decent, but can also be used to indicate something really great. I don't want visitors to consider the section to be endorsements.

I think "Notable" is slightly better, but I'm not getting anything I like a lot from the thesaurus. "Popular" is correct but misleading.

🤷 Push it and change later if we think of something? I don't want to hold it up for semantics.

PR #433, opened 2026-06-22
Restore card focus on back

kaste:
Remember keyboard-activated card links and restore focus to the same card when the user returns with browser history.

In a nutshell:  Use the arrows to move between cards, `enter` to visit the page/package, `alt+left` (aka `back`) to return to the homepage *but* the previous selected card is selected again.


michaelblyons:
I hadn't been using Alt-Left, but I think I will now.

The keyboard interface for PCR is really nice.

PR #435, opened 2026-06-22
Split Angular label aliases

kaste:
Keep angular2 mapped to angular, but canonicalize angular.js to angularjs per the review follow-up.

### DIFF
--- a/label-aliases.json
+++ b/label-aliases.json
@@ -7,7 +7,9 @@
     "amxmodx"
   ],
   "angular": [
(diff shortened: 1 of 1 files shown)


kaste:
@michaelblyons 👀 or not worth it?

michaelblyons:
Yeah, let's do it, and I'll see if I can convince people to tag some of the packages.

Side note: AngularJS is a *trash* language compared to Angular.

PR #439, opened 2026-06-23
Bundle js

kaste:
Add esbuild and a bundled filter for opt-in production script URLs.
Templates keep source URLs in development, while production points
bundled entries at the static hash bundle directory.

Bundle the main route entries and global theme/key scripts without
code splitting so output filenames remain predictable.


kaste:
This worked locally but still 🤞

PR #440, opened 2026-06-24
Fix vertical alignment of icons

BenjaminSchaaf:
Before:

Screenshot from 2026-06-24 18-05-59

After:

Screenshot from 2026-06-24 18-06-13

### DIFF
--- a/_includes/footer.njk
+++ b/_includes/footer.njk
@@ -15,7 +15,7 @@
 
   <nav>
     <a id="footer-links" href="https://github.com/sublimehq/package_control_channel">Submit your own package</a>
-    <a class="github-link" href="https://github.com/packagecontrol/thecrawl/tree/gh-pages">
+    <a class="github-link inline-icon" href="https://github.com/packagecontrol/thecrawl/tree/gh-pages">
       {{ util.icon('octocat') }}
       <span class="screenreader-only">This website on GitHub</span>
     </a>
--- a/_includes/header.njk
+++ b/_includes/header.njk
@@ -39,7 +39,7 @@
     <nav>
       {% if isPackagePage %}
         <a class="search-toggle" href="/#search-field">
-          {{ util.icon('search') }}
+          <span class="inline-icon">{{ util.icon('search') }}</span>
           <span class="screenreader-only">Search</span>
         </a>
       {% endif %}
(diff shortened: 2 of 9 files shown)


kaste:
This seems wrong as there is no `vertical-align: center;`

image

On my machine after looks like

image

michaelblyons:
Does it look okay if you change `center` to `middle`?

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align

BenjaminSchaaf:
Ah, `vertical-align: center` is a firefox-ism, and doesn't do the same as `middle`. Seems the modern approach is to use flexbox.

BenjaminSchaaf:
I've pushed a fix for this.

braver:
Icon sizing on the labels overview page needs a little help 😄 

Scherm­afbeelding 2026-07-01 om 12 37 59

This might be the patch:

```diff
diff --git a/labels.njk b/labels.njk
index e95408b9..1f73a3c0 100644
--- a/labels.njk
+++ b/labels.njk
@@ -55,7 +55,7 @@ page_type: labels
           {% set icon_id = item.key | label_icon_id %}
           {% if icon_id %}
             {% set tint = item.key | label_icon_tint %}
-            <svg class="{{ (' label-icon--' ~ tint) if tint }}" aria-hidden="true">
+            <svg class="inline-icon {{ (' label-icon--' ~ tint) if tint }}" aria-hidden="true">
               <use href="/{{ 'static/label-icons.svg' | bust }}#{{ icon_id }}"></use>
             </svg>
           {% endif %}
```

In general icons in labels are now smaller than before (they were 16x16 pixels, which is preferable as that's typically the size they're designed for, meaning you get partial-pixel line widths etc):

before
Scherm­afbeelding 2026-07-01 om 12 38 59

after
Scherm­afbeelding 2026-07-01 om 12 39 03


```css
    vertical-align: -0.125em;
```

This doesn't seem to actually do anything on my machine (Chromium). I don't think it should either, the inline-icons are in a `.button`, which is a flex box.

The PR does improve the alignment of the "stats" icons though, it just seems to have a bit of fallout elsewhere.

kaste:
This is -- minus bugs -- essentially a change from vertical-align: -1px to -2px.  As such it is not a "Fix" because the previous value was manually dialed in but a different rendering.  This looks good on high DPI but not on low DPI screens where the change is from 

image

to

image

t.i. -2px doesn't look "more" centered.

braver:
Maybe a `@media (min-resolution: 2x) ` is needed so that it can be perfect in both low and hi-res displays?

PR #442, opened 2026-06-24
Improve readability of light-mode theme toggle

BenjaminSchaaf:
The big yellow gradient makes the icon impossible to read when the button is focused.

Before:

image

After:

image

### DIFF
--- a/static/style/theme-toggle.css
+++ b/static/style/theme-toggle.css
@@ -50,7 +50,7 @@
       .light {
         outline: 1px solid rgba(255, 190, 16, 0);
         box-shadow: inset 1px 1px 10px -1px #ffe73500, 0 0 0 rgba(253, 225, 0, 0);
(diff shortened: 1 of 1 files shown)


kaste:
>  impossible to read

How do you actually read the before and the after state?

Right now, your patch renders

image

btw

kaste:
I removed the easter egg, which I did I think last summer with my younger daughter; actually she did and I just clutched it so that it holds ...   a bit.

Deployed is now a new, simplified baseline and for which we can find new easter eggs.

kaste:
The before state was a sun, completely possible to read IMO:

image

BenjaminSchaaf:
I certainly prefer what's been merged now. Fwiw the gradient transparently overlaying the icon made it look like a mistakenly implemented glow rather than a sun - it would have worked better if the icon wasn't visible.

PR #443, opened 2026-06-24
Don't use border and outline for the same visual element

BenjaminSchaaf:
When focused the search field has an outline, but the border is also being used to show this outline. This causes some odd rendering issues:
* Both Firefox and Chrome have a gap in the corners between border and outline
* Firefox sometimes has gap on edges between the border and outline

To fix this we can use an outline of 2px and overlay it by 1px, instead of using the border.

Before:

image

After:

image

### DIFF
--- a/static/style/search.css
+++ b/static/style/search.css
@@ -156,8 +156,8 @@ select {
 
   &:has(input:focus),
   &:has(.clear-btn:focus) {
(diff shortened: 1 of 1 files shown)


kaste:
Makes sense.  I don't have that visual quirk with my browser.

PR #445, opened 2026-06-24
Always have sourcemaps enabled

BenjaminSchaaf:
There's no reason to limit debugging in production, there's nothing to hide in the source code.


kaste:
That's okay if you like it that way.

PR #446, opened 2026-06-24
Fix flashing on icons when transitioning themes in firefox

BenjaminSchaaf:
Firefox has a bug where it doesn't inherit animations on backgrounds. There seems to be no reason to inherit the background anyway, so the workaround is to just remove it.


kaste:
👍 ;  doesn't regress on my machine

PR #449, opened 2026-06-24
Optimize for high-usage label icons

kaste:
... but proper and dynamic.

Also, for neither "apple" nor "mac" we have icons.  Remove, the c family because "Jon finds them awful".


kaste:
Measured against `static/label-icons.svg` before vs after removing the C-family icons plus low-usage icons:

| sprite | raw | gzip | brotli | icons |
|---|---:|---:|---:|---:|
| before | 200.5 kB | 81.8 kB | 71.6 kB | 141 |
| after | 113.4 kB | 45.4 kB | 39.9 kB | 103 |
| saved | **87.1 kB** | **36.5 kB** | **31.7 kB** | **38** |

Breakdown:

- C-family only: ~2.8 kB raw, ~0.8 kB gzip, ~0.6 kB brotli
- Low-usage icons: ~84.2 kB raw, ~35.7 kB gzip, ~31.1 kB brotli

So the low-usage pruning is basically all of the win.


Assuming the sprite is downloaded and not already cached:

| network | brotli saved time | gzip saved time |
|---|---:|---:|
| Slow 3G-ish, 400 kbps | ~635 ms | ~730 ms |
| Lighthouse mobile / Fast 3G, 1.6 Mbps | ~160 ms | ~180 ms |
| Decent 4G, 9 Mbps | ~28 ms | ~32 ms |

- This does not save a network round trip unless the request disappears entirely.
- Cached repeat visits save nothing on network transfer, obviously.

braver:
Any somewhat normal website includes several images of hundreds of kilobytes each. I don't really understand why we need to save a few kilobytes here. The "search index" is so much larger, it's much easier to win milliseconds there by tweaking how or when that loads.

Alternatively, the icons file is loaded completely regardless of what icons you're looking at right now. Instead, we could consider inlining the svg content into the markup instead of referencing elements in the larger file. That way you only load what you're rendering at that point. ([random internet link](https://medium.com/@brettdewoody/inlining-svgs-in-eleventy-cffb1114e7b) about this approach):

Scherm­afbeelding 2026-07-01 om 12 23 15

michaelblyons:
> Any somewhat normal website includes several images of hundreds of kilobytes each. I don't really understand why we need to save a few kilobytes here. The "search index" is so much larger, it's much easier to win milliseconds there by tweaking how or when that loads.

Agree.

> Alternatively, the icons file is loaded completely regardless of what icons you're looking at right now. Instead, we could consider inlining the svg content into the markup instead of referencing elements in the larger file. That way you only load what you're rendering at that point. ([random internet link](https://medium.com/@brettdewoody/inlining-svgs-in-eleventy-cffb1114e7b) about this approach)

I don't know that this is best. You save the cost of the unvisited icons, but you now load the SVG bytes every place it appears in the file. And you don't get to cache the bytes across different pages with the same icon (let alone between visits where the HTML updated but the icon doesn't).

Can eleventy provide each icon as a file? You add the bytes of overhead for transferring each file, but you get to cache the files again.

braver:
We don’t cache any of the other the words or bits markup either. You’re talking literally about bytes at that point.

> Can eleventy provide each icon as a file?

They might even start out as individual files? Not sure what the build pipeline looks like for this.

kaste:
If this is just a fight between first load/most commonly used icons (as I read this), we just split the icons sprite into 2 (or n but that's really overdone then), and ship 100 icons on the front page and the rest by their first usage.  (We actually ship on first usage anyway, but the homepage almost always has icons.)

kaste:
```
 ┌─────────────────────┬───────┬───────────────┬───────────────────────────────┬───────────────────────┬───────────────────────┐
 │ Primary rule        │ Icons │ Primary       │ Packages with icons fully     │ First secondary:      │ First secondary:      │
 │                     │       │ Brotli        │ covered                       │ Newest                │ Updated               │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 1 +         │ 105   │ 39.9 KiB      │ 96.4%                         │ page 24               │ page 16               │
 │ homepage            │       │               │                               │                       │                       │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 1 + 6       │ 107   │ 41.0 KiB      │ 96.6%                         │ page 36               │ page 47               │
 │ months              │       │               │                               │                       │                       │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 1 + 12      │ 107   │ 41.0 KiB      │ 96.6%                         │ page 36               │ page 47               │
 │ months              │       │               │                               │                       │                       │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 2 + 6       │ 90    │ 35.8 KiB      │ 93.2%                         │ page 32               │ page 46               │
 │ months              │       │               │                               │                       │                       │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 2 + 12      │ 90    │ 35.8 KiB      │ 93.2%                         │ page 32               │ page 46               │
 │ months              │       │               │                               │                       │                       │
 ├─────────────────────┼───────┼───────────────┼───────────────────────────────┼───────────────────────┼───────────────────────┤
 │ usage > 3 + 12      │ 81    │ 32.0 KiB      │ 90.1%                         │ page 32               │ page 46               │
 │ months              │       │               │                               │                       │                       │
 └─────────────────────┴───────┴───────────────┴───────────────────────────────┴───────────────────────┴───────────────────────┘
```

So usage > 2 seems like a better fit (or even 3), as we still had 93% of all packages covered by that first sprite file but only 35KiB download size.  

+ homepage = used on the homepage
+ 6/12 months = used on a package that changed or was added in that time period

Issue #429, opened 2026-06-20
Labels containing the text of the highlighted ones are also bold


michaelblyons:
https://packages.sublimetext.io/labels?q=theme

kaste:
I guess bravers simplicity hits us :grin:

```
  &.label {
    &[href*='language%20syntax'],
    &[href*='snippets'],
    &[href*='linting'],
    &[href*='auto-complete'],
    &[href*='color%20scheme'],
    &[href*='theme'] {
      color: var(--foreground-1);
      font-weight: bold;
    }
  }
```

kaste:
I doesn't look bad, however.

<img width="861" height="451" alt="Image" src="https://github.com/user-attachments/assets/021b711b-c064-4aa5-8bcc-b44d479de42d" />

Still some kind of grouping.

braver:
Hey, if you're going to blame me just `@` me! 😉 

Ok, maybe this is also a simple solution but these 3 should not be labels:

<img width="473" height="88" alt="Image" src="https://github.com/user-attachments/assets/d571a2bd-c843-495a-bf01-6ae2e383980f" />

"hadar" and "material" are just the names of the packages not a specific kind of theme. Theme packages should have the `theme` label singular as per [the rules](https://docs.sublimetext.io/guide/package-control/submitting.html#what-labels-to-use). I would (try to) not accept such labels in new package submissions. 

We could still make the implementation less simplistic, but in this case I'd like to fix it in the registry itself.

kaste:
I don't think we need to tackle anything here.  It looks good to me.

PR #457, opened 2026-07-01
the common label is "material"

braver:
From https://github.com/packagecontrol/thecrawl/issues/429 @michaelblyons

Related work in the live registry/channel: https://github.com/sublimehq/package_control_channel/pull/9463

The common label for packages related to Material Design is "material", not variations on "material theme", "material design" etc. This PR updates 2 dead packages so they don't pollute the dataset.

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -44827,8 +44827,7 @@
       "name": "Material Theme - Appbar",
       "labels": [
         "theme",
-        "material design",
-        "material theme"
+        "material"
       ],
       "first_seen": "2016-01-07T23:56:43Z",
       "removed": "2021-12-19T18:07:48Z"
@@ -44838,7 +44837,7 @@
       "labels": [
         "theme",
         "color scheme",
-        "material design"
+        "material"
       ],
       "first_seen": "2015-06-25T19:00:09Z",
       "removed": "2021-12-19T18:07:50Z"


michaelblyons (approved):

michaelblyons:
Obsolete? The sync bot seems to have taken care of it.

braver:
Not sure what you mean? This PR targets 2 packages that are not in package_control_channel anymore, so there is nothing to sync. This is just their "tombstones" that are actually still on the site and pollute the labels list:

Screenshot 2026-07-03 at 19 40 04

michaelblyons:
Oops. My mistake. :blush: I thought this was a strict duplicate of https://github.com/sublimehq/package_control_channel/pull/9463

kaste:
Admitted, I'm in the "don't change the social artifacts team", but why should we do this for dead packages?

braver:
Well we did some useful work in https://github.com/sublimehq/package_control_channel/pull/9463 to clean up some ambiguity or duplication with some labels. I'm not really on team "let's clean up the labels", it's a bit like going through twitter and telling people to be more consistent with their hashtags. They're just "social artifacts" as you say. Besides the handful important ones they don't really matter. This PR is just to finish what we started, but I'm 100% ok with not doing it as well. We can also just not fuss over labels too much, especially on the dead packages.

PR #461, opened 2026-07-21
Avoid re-rendering readmes

BenjaminSchaaf:
The readme rendering is slow and memory intensive, but there's no reason to redo all that work every time. We now store a hash of the text used to render the readme inside readmes_renderes.json to avoid re-rendering.

When no readme has changed this bring rendering time from ~30s to ~700ms and of course significant memory savings.

FWIW I did a quick and dirty rewrite in rust and a full uncached render takes just 2s and ~50MB of RAM.


kaste:
Cherry-picked.  The main problem with this commit was that it only considered the source text to invalidate the cache, but the other side, the renderer, can also change.  Also the root cause should be fixed as we really run out-of-memory here quickly.  With the root fix applied, the question was if we should cache at all as caches can become stale but I still went with it as it really seems natural to do so.        

The follow-up is #462

PR #465, opened 2026-07-22
Queue searches while index loads

### LINKED ISSUE #460: Search not working on new page load
When the page is first loaded, if you enter a search term before the search database is loaded no search is performed. You need to edit the search field after the database has loaded for search to work. In the case where the database has not loaded it should wait until it has and then perform the search immediately, rather than dropping that event.

kaste:
Keep the search field editable and use its current value as the pending query. Show an accessible loading notice only after a query is entered, then run that query once the index becomes available.

Preserve failure feedback and allow clearing the field to cancel a pending search.

Fixes #460.

### DIFF
--- a/_includes/header.njk
+++ b/_includes/header.njk
@@ -34,6 +34,7 @@
   <div class="right-side">
     {% if isPackagePage %}
       {{ util.form(collections.packages | length, false) }}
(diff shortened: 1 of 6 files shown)


kaste:
image
image

PR #466, opened 2026-07-23
Remove R from Package Control branding

kaste:
Remove the R suffix from the site heading and About introduction, then drop the now-unused hot text style.


michaelblyons:
It's still in the `<title>`.

kaste:
🤣

PR #482, opened 2026-08-17
Generalize package source deep links

kaste:
Generate a source location model from shallow checkouts of GitHub repositories with more than five living packages. This provides deep links for 5,105 of 5,171 living packages (98.7%).

Teach Eleventy to consume the generated model while retaining fallback links when the model or an exact package location is unavailable.

--

(IMO) The source deep links are so great that it is worth to do it right.  First implementation was basically ad-hoc and a bit awkward, so we generalize here and make it look sane.

This is introduces (thus) also a new make step as the links are now generated "as a sidecar" before eleventy runs.  (Really, the first implementation did this within eleventy.config on every dev run. :grin:)  

There is a `SOURCE_THRESHOLD` knob here so we don't rely on hard-coded paths and react automatcially to changes in the dataset.  Almost 99% coverage as of now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ _site/*
 ######################
 .AFileIcon/
 .package_control_channel/
+.source_repositories/
+source-model.json
 channel.json
 logs.json
 readmes.json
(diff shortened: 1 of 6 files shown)


kaste:
Examples?  All LSP and SublimeLinter packages now have deep links.

PR #490, opened 2026-09-05
Install chart iteration

kaste:
* Removes the weekly dots on the upgrade line.  These dots were at some point hover targets, but they're not anymore so we can remove them just keeping the release dots.
* Dampens the line on the upper and lower bounds.  This came out from a thinking error as I wanted a strict limit at the x-axis (to keep the lines *inside* the chart). ... but I really like that math and it is simpler as the straight lower bound limiter too.  Just less curvy, childish-playful.
* We always had dailies for the running month or so but we never used it, essentially because of the scaling math.  (dailies has a different scale than weeklies, so what to do)  There are several solution to the problem; here we take a trailing window (day of the daily data plus 6 days towards the history).  We have nice transition from weekly to daily data, overall I think this looks good and expressive.

Before 

image

After

image

### DIFF
--- a/eleventy.config.mjs
+++ b/eleventy.config.mjs
@@ -33,6 +33,8 @@ const LABEL_ICON_MINIMUM_USAGE = 3
 const LABEL_ICON_RECENT_WINDOW_DAYS = 365
 // The main chart width is 53 weeks; + 1 for a possible continuation line
 const INSTALL_CHART_WEEKS = 53 + 1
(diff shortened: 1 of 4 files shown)


kaste:
Maybe a follow up PR which exactly defines and designs the transition between weekly and daily line might come.  That just needs a step back.

E.g. (a) dailies must start on week boundaries (b) dailies start at the first release dot (in the available range (~30 days) but at least on or before the running week.

PR #491, opened 2026-09-05
Remove local status page

kaste:
Link to the crawler's status page from the footer and remove the local status implementation, assets, tests, build support, and log data flow.

image

1. The status page was not copied over by sublimehq.  
2. thecrawl's status has already more features, esp. searching the notes/logs and seeing when a specific package has been "touched" by the crawler.  While obviously not impossible or even difficult to port to anything else, the implementation relies on GitHub's infrastructure as it reflects the Status of that infrastructure.


michaelblyons:
Is PCR status still going to work after this change? I see you made an absolute link to it.

kaste:
PCR status is now at https://packagecontrol.github.io/thecrawl/status/.  That's the idea at least.

kaste:
@michaelblyons A status page is a bit tied to the actual infrastructure.  You can port the new features easily, eg. 

image

but only if you know the infrastructure.  I think the "search notes" is worth it.

Also https://packagecontrol.github.io/thecrawl/status/?run_id=33566759152&q=mise

image

That's just good stuff IMO.

PR #498, opened 2026-09-10
Build a package graveyard

kaste:
Render all tombstones as a dense dated archive with stable deep links.

Move nine-month-old tombstones out of the regular package collection.
Route package succession links to either live package pages or graveyard
entries.

In search results, list possible tombstone matches as an addendum.

Keep Sublime Text 2 packages on their full pages but show their search
matches also as a separate addendum.

E.g de-emphasize ST2 packages

image

image

image

### DIFF
--- a/404.njk
+++ b/404.njk
@@ -2,6 +2,7 @@
 layout: layout.njk
 permalink: 404.html
 title: 404
+isNotFoundPage: true
 eleventyExcludeFromCollections: true
 ---
 
@@ -17,10 +18,7 @@ eleventyExcludeFromCollections: true
   <p><a id="search-btn" style="display:inline-block" class="button" href="/#search-field">Search for something</a></p>
 </section>
 
-{#
-  Provide a package name map for case-insensitive lookup on 404s.
-  Map lowercased name -> correct-cased name.
-#}
+{# Map lowercased package names to their canonical case for 404 suggestions. #}
 <script>
   (function(){
     var map = {
--- a/404.njk
+++ b/404.njk
@@ -18,16 +18,13 @@ eleventyExcludeFromCollections: true
   <p><a id="search-btn" style="display:inline-block" class="button" href="/#search-field">Search for something</a></p>
 </section>
 
-{# Map lowercased package names to their canonical case for 404 suggestions. #}
+{# Package names for case-insensitive lookup and typo suggestions. #}
 <script>
-  (function(){
-    var map = {
-{%- for p in collections.packages -%}
-      "{{ (p.name | string) | replace('\\', '\\\\') | replace('"','\\"') | lower }}": "{{ (p.name | string) | replace('\\', '\\\\') | replace('"','\\"') }}"{% if not loop.last %},{% endif -%}
+  window.__PKG_NAMES = [
+{%- for pkg in collections.packages -%}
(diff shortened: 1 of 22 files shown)


kaste:
That's actually all I had in my notes... I'm done.

braver:
Cool 👌🏻 Some questions about the visuals:

1. What is the red plus sign for?
Scherm­afbeelding 2026-09-20 om 21 11 40

Hmm ... also, there's a plus here:
Scherm­afbeelding 2026-09-20 om 21 12 53

In your screenshots the horizontal bar on the plus looks misaligned a bit... almost like `✝` (latin cross). Perhaps that's the sign you're looking for? From what I can tell it looks like you're using an actual plus sign though. 🤔 

2. Why is the H1 in IBM Plex Mono rather than the regular Lato? I kind of like it, almost has a gothic feel to it somehow 🪦🦇. Just wondering if that was intentional.

kaste:
Yes on both.

1. I first tried the latin cross.  But that was just too narrow, thin and long
in our font.  They almost put a Jesus on it.  So a greek cross.  And the Plus
sign on my low DPI actually looked already good because with the small font I
already get aliasing.  And then I probably forgot about.  A hand-made SVG,
based on a greek cross just with the patibulum one pixel up but same width
would nail(:grin:) it.

2. That's literally what I did: in google fonts, typed Graveyard and clicked 
Gothic.  Our monospace had already some of it and it comes free, no separate 
download.  (Although one can reduce the font to just the used letters; graveyd.)

The tombstones look more like this on my screen.

image

Yours is huge.

braver:
> Yours is huge.

🤔... 

Yeah, that's just whatever GitHub did to my `@2x` resolution screenshot. You've got some weird hinting going on. 

Anyway, this what you're looking for?

Screenshot 2026-09-21 at 11 18 33

```html
<svg width="8" height="8" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" xmlns:affinity="https://www.affinity.studio/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;">
    <g transform="matrix(1,0,0,1.090909,-12,-9.454545)">
        <path d="M19,15.083L12,15.083L12,13.25L19,13.25L19,8.667L21,8.667L21,13.25L28,13.25L28,15.083L21,15.083L21,23.333L19,23.333L19,15.083Z"/>
    </g>
</svg>

```

kaste:
Hm...  this is what ✝ actually looks like with the current design

image
image

So that is actually good (better than current anyway)... The feature took quite a while so I may have tested the actual latin cross only with different fonts, font-weights, font-sizes  ...

On the graveyard it looks quite authentic.  The right side on the compact card ... we have options here now

image
image

Eg. we can maybe also just omit the right side.  This was different in an earlier version, where the RIPs were still embedded within the normal results. But now they have their own nice section.   

Right now in the moment, I would just go with the ✝.  
Maybe even in sad black:
image

braver:
> Eg. we can maybe also just omit the right side

I would go with that. Unless it can be used to reveal something (like dates on hover) it doesn't add anything: in this section they're all dead.

> Hm... this is what ✝ actually looks like with the current design

Looks good actually 👍🏻

kaste:
We can of course just add the standard tooltip and show the actual date.  --> #499

PR #499, opened 2026-09-21
Use crosses for removed packages

kaste:
Replace plus markers with Latin crosses on graveyard cards and compact search results. Show the correctly formatted removal date when hovering the search result marker.


braver:
🎉

